Issue #8, opened 2016-02-21
Remove unneeded files and folders from repository


bittner:
There are some files and folders which are not supposed to be under version control, e.g.

- build files and folders, such as `*.egg-info` (they can be regenerated anytime)
- Python byte code files (`*.pyc`) and Python3's `__pycache__` folder (generated by Python, may produce unpleasant side-effects when distributed to other systems)
- editor configuration, such as `transcrypt.npp` (is highly dependent on your local system, and changes frequently)
- ... (this list is not complete)

I suggest you remove them from the repository, and add a `.gitignore` file with appropriate definitions to avoid adding any such files by mistake in future. See GitHub's [gitignore](https://github.com/github/gitignore) repository for useful samples.

JdeH:
Thanks

I currently use a .gitignore, but its incomplete. Indeed the stuff you mention should be omitted.
I'll look into your .gitignore before the next upload and either use it as a whole or merge it with mine.

Jacques

JdeH:
Somewhere along the way the .gitignore file was accidentally deleted.
I'll add it again from a previous version.

JdeH:
Should be solved now.
Feel free to reopen issue if not.

PR #9, opened 2016-02-22
Created a .gitignore

techdragon:
Partially addresses #8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,73 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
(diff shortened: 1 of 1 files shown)


JdeH:
Thanks, 

I'll compare that to my own gitignore a.s.a.p. and either merge the two or just use yours.

Jacques

JdeH:
Thanks

I'll look into your .gitignore before the next upload and either merge it with mine or use it as a whole.

Jacques

JdeH:
Somewhere along the way the .gitignore file was accidentally deleted.
I'll add it again from a previous version.

JdeH:
Should be allright now.

PR #56, opened 2016-06-08
Add string slicing

pierrejean-coudert:
Update builtin.mod.js : add String.prototype.getslice

Allow string slicing:

```
s='abcdefghij'
print(s[2:3])
print(s[:3])
print(s[2:])
print(s[::2])
```


JdeH:
Thanks!

PR #59, opened 2016-06-13
add map & filter builtin functions

pierrejean-coudert:


JdeH:
Hi,

To make this work some code had to be added to __builtin__.mod.js:

	var sorted = __all__.sorted;     # Already existing line
	var map = __all__.map;           # New line
	var filter = __all__.filter;           # New line

The reason for this is that __standard__.mod.js is an "in line" module, it isn't imported in the usual way but is in Transcrypts global namespace by default, bypassing the import mechanism. By adding the two lines above to __builtin__.mod.js, your new functions become part of the builtins, even if they are programmed in Python in __standard__.py rather than directly in JavaScript in __builtin__.mod.

I've also added a testcase to the autotest in the div_pulls module. Note that for CPython equivalence, explict conversion to 'list' is done, since otherwise in CPython the result would be an iterator rather than a list, which Transcrypt doesn't yet support.

Kind regards
Jacques

pierrejean-coudert:
Thanks for the explanation.

Would it be possible to exclude from the commits all the build generated files ? It would really ease the reading of diffs.

One more thing, would you mind if we convert indentation from Tabs to 4 Spaces (as in PEP8) ?

JdeH:
First point: Indeed including all the generated files is very unhandy. I have not yet found a simple automated way to exclude all of those, since some .js files should be retained (js only modules). Come to think of it, probably just checking the date would do... I'll look into it.

Second point: As for PEP8 and the four spaces, there have been some previous discussions, also on the use of Camel Case for variables. Since Transcrypt is part of a larger multi-language project (C++, Python, JavaScript) with team coding standards, I will preserve the tabs in the core modules, together with camel case for variables etc.

However in any examples, I will preserve whatever whitespace type is choosen. E.g. in your React demo, I've just preserved the spaces.

The same will hold for libraries. If you supply a library where you use 4 tabs, I will preserve them in the repo. If I add anything to your library, I will use spaces, not tabs.

This is as close as I can get to PEP8 in my practical situation.

JdeH:
Many generated files removed.
Docs still give some garbage.

PR #77, opened 2016-07-11
Riotjs demo

AXGKl:
Added a Demo for RiotJS. Would be glad if you could consider including it into Transcrypt.

And... DANKEWELL FOR THIS *AMAZING* TOOL!!

### DIFF
--- a/transcrypt/demos/riot_demo/.gitignore
+++ b/transcrypt/demos/riot_demo/.gitignore
@@ -1,0 +1,1 @@
+__javascript__/*
--- a/transcrypt/demos/riot_demo/README.md
+++ b/transcrypt/demos/riot_demo/README.md
@@ -1,6 +1,6 @@
 # RiotJs Demo
 
-Riot is a 'one way binding' component framework, comparable to React but with
+[Riot](https://github.com/riot/riot) is a 'one way binding' component framework, comparable to React but with
 a far simpler API.
 
 If features a template engine, lifecycle events and tag local namespacing,
--- a/transcrypt/demos/riot_demo/README.md
+++ b/transcrypt/demos/riot_demo/README.md
@@ -1,0 +1,39 @@
+# RiotJs Demo
+
+Riot is a 'one way binding' component framework, comparable to React but with
+a far simpler API.
+
+If features a template engine, lifecycle events and tag local namespacing,
+all based on HTML5, js and css only.
+
+This demo shows how to write and mount a Riot Tag (sample_tag2.py) in pure Python,
+inheriting from a base class which takes care of binding all variables to the tag
+instance, so that existing frameworks using riot js based tags can be converted
+to Transcrypt, one by one with no(?) effort outside the tag source itself.
+The base class is using some direct JS.
+
+## Usage
+
+You can add js logic also outside transcrypt js pragmas, see the riot.tag2
+function in the html file.
+The transcrypt tag is available via `this.py_obj` from the riot tag and vice
+versa via `self.riot_tag`.
+
+
+## Transpiling and Running
+
+Run the demo with no switches on transcrypt and serve `riot_demo.html`
+
+## Misc
+
+For reference reasons I added a classic JS based tag, including a compiled
+version. The compilation can also be done (fast) in the browser for such
+classic tags.
+
+
+## Disclaimer:
+
+I'm for sure no JS expert, there might be a far more elegant way to
+have riot's 'this' and Transcrypt's self in sync, e.g. via JS 'inheritance'
+mechanics - or maybe using Transcrypt's object extension methods itself.
+
--- a/transcrypt/demos/riot_demo/riot_demo.html
+++ b/transcrypt/demos/riot_demo/riot_demo.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+
+  <head>
+    <meta charset="UTF-8" />
+    <title>Hello PyRiot ;-)</title>
+    <!-- for transcrypted version we do not need the compiler.
+         for the classic version you can use riot+compiler -->
+    <script src="https://cdn.jsdelivr.net/riot/2.5/riot.js"></script>
+	<style>
+        body {font-family:arial;font-size:30px;padding:50px;}
+		h1 {font-size:50px;color:#0000ff;}
+	</style>
+  </head>
+
+  <body>
+
+  <!-- classic riot (compiled on server, could be done in browser)         -->
+        <sample id="s1"></sample>
+        <script src="tags/sample.js"></script>
+        <script>riot.mount('sample')</script>
+
+
+  <!-- transcrypted version (transpiled on server, using python)           -->
+        <sample2 id='s2' label="nr1"></sample2>
+        <mp2     id='s3' label="nr2"></mp2>
+        <!-- the transpiled tag's js -->
+        <script src="__javascript__/sample2_tag.js"></script>
+        <script>
+            // register
+            var cls = sample2_tag.Sample2
+            riot.tag2('sample2', cls.template, cls.style, '', function(opts) {
+                new cls(this, opts)})
+            document.t1 = riot.mount('sample2')[0].update()
+            document.t2 = riot.mount('#s3', 'sample2')[0]
+            document.t2.update()
+        </script>
+  </body>
+</html>
--- a/transcrypt/demos/riot_demo/riot_tag.py
+++ b/transcrypt/demos/riot_demo/riot_tag.py
@@ -1,0 +1,89 @@
+# Parent Class for a Transcrypt Riot Tag
+#
+# This binds the namespace of a riot tag at before-mount event 100% to that of
+# the a transcrypt instance, except members which begin with an underscore, those
+# are private to the transcrypt object.
+#
+# The 4 riot lifecycle events are bound to overwritable python functions.
+#
+# Immutables (strings, ints, ...) are bound to property functions within the tag,
+# so the templates work, based on state in the transcrypt tag.
+# State can be changed in the riot tag as well but take care to not create new
+# references - you won't find them in the Transcrypt tag.
+#
+#
+# Best Practices:
+#  - mutate state only in the transcrypt tag.
+#  - declare all variables so they are bound into the riot tag
+#  - IF you declare new variables to be used in templates, run
+#    self.bind_vars(self.riot_tag)
+# TODO: docstring format not accepted by the transpiler, strange.
+
+__author__ = "Gunther Klessinger, gk@axiros.com, Germany"
+__date__ = "$Date: 2016-07-11 02:17:51 +0100 $"
+
+class RiotTag:
+    """
+    taking care for extending the riot tag obj with
+    functions and immutable(!) properties of derivations of us
+    See counter.
+    """
+    debug = None
+    # placeholders:
+    template = '<h1>it worx</h1>'
+    style = ''
+    node_name = '<no node>'
+    opts = None
+    def __init__(self, tag, opts):
+        # opts into the python instance, why not:
+        self.opts = opts
+        self._setup_tag(tag)
+
+    def _setup_tag(self, tag):
+        # keeping mutual refs
+        tag.py_obj = self
+        self.riot_tag = tag
+        # making the event system call self's methods:
+        __pragma__('js', '{}', '''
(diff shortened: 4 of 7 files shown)


JdeH:
Hi Gunter,

Thank you for your contribution! I've added it to the demo's.
As you have noted, the 'super' keyword has not yet been implemented,
currently you have to call parent class constructors explicitly via the name of the parent class.
Leaving 'super' out was part of the strategy to keep Transcrypt small.
There's no real hard criterion on what should be in or out, 'super' is on the edge.
Do you have a strong opinion on wether it should be in or out?

Kind regards
Jacques

AXGKl:
Hi Jacques,

First of all: Sorry for changing my branch name to axgk plus an update of the riot demo, with conflicts, forgot that I can't resolve the conflict myself (we use mercurial, keep making mistakes in [git](https://git-man-page-generator.lokaltog.net/)). Just update the riot folder clean if you are ok with it and sorry again.

I changed the branch name since I find TC (ok with that abbr?)

1. *incredibly well designed* (especially the always available fallback to js from module to statement level is absolutely amazing and reassuring that you'll never hit the wall, mid ways in a project. Also the module organization or the autotester...)
2.  coded so well, regarding expandability, that I think it does have game changing potential in certain important areas. I think any python intermediate and js beginner is able to understand the basic design and where it is implemented pretty quick. And that is key, imho regarding TC's future (and I did try a bit to understand other tools before). 

So I changed the branch name to a more generic one, since I might propose more stuff to you, also outside the riotjs folder, here and there. Unless you stop me from doing so, that is.

----

Regarding `super`: No problem at all, I remember vaguely I did try the explicit parent class but I made mistake obviously, came to the conclusion that it also did not work. But it does and I adapted the demo. Guess it was too late and me = too excited ;-)  Actually, I like it better anyway, regarding the one big selling point for Python and that is, imho, clean code, understandable on a high level w/o the immediate need to understand the details.

----

Regarding the adding of that little [`dir()` function](https://github.com/JdeH/Transcrypt/pull/80/commits/c60bd5e4ef072c896917f5e6902510c5b9109b2d):
Whats your policy here?
The output of the runtimes for sure does not match for the "dunder" variables of js objects versus those of py. Just need to know your general position here, so that I know how to organize myself, regarding stuff I'm missing.
E.g. I added a few [TODOs](https://github.com/AXGKl/Transcrypt/blob/30e29864caa3a3a0476a8a8e672d2b57a2003ee0/transcrypt/demos/riot_demo/color.py), which I guess I could fix as well sooner or later. I don't want to bug you with core pull requests if you say that thats the clean common denominator, everything else must stay outside.
You write bloat avoidance is of utmost prio and I'm 100% in line with that, thats why dir() is like it is ;-)

----

Ok, enough for now, have a nice weekend!
g.

JdeH:
Hi Gunter

First of all about Transcrypt in general. I too have the feeling that it has considerable potential. I think the main hurdle is to allow developers to gain trust that this project has continuity and doing commerial work in it is a relatively safe investment. This means Trancrypt should try to stay close to Python to allow some degree of second sourcing. Also it should be coded and documented in such a way that a company using it principle can maintain and update the code if large projects depend upon it. I hope to stay with Transcrypt for a long time to keep it lean and on track, but I don't expect larger companies to want to become dependent on one person they don't control or even know.

I've thought quite a lot about the future of this project before I started investing so much energy in it after a modest start. Two questions had to be answered:

Q1. Won't the developments with regard to JS itself not make such a project redundant in only say two years.
Q2. How long before CPython iteself runs in the browser loading at acceptable speed.

The answers I gave to myself were:

A1. JS will keep growing and its community will be even more massive, for sure. But its programming paradigm differs fundamentally from Python's.  For someone used to class based OO it'll probably always feel messy and artificial, no matter how much paint is applied to the surface to hide its prototype based, typeless guts. Because of JS wide acceptance, it's too late for ECMA to throw certain things away, as far as a committee (as opposed to a person) can ever do that. Since Python has such a prominent place in education and in the scientific community, there will remain a large group of developers that will feel better at home with Pythons clean design and syntax than with JS. 

A2. PyPyJs is a very solid attempt to bring plain CPython to the browser. While this fundamental approach certainly has a lot of appeal, and certainly has my sympathy, there are some problems with it that from practical experience I know could prove quite persistent. First of all there's the problem of page load speed. Looking at it from a distance it's  tempting to say: well, since browsers get better at caching things locally, why wouldn't the whole CPython interpreter by cached in the end, as a kind of informal browser add-in. This seemed quite plausible until I contemplated the fate of Java with the advent of C#, the fate of Flash with the advent of iOS and the fate of OS2 with the advent of Windows XP. Often technologically superior solutions bite the dust for commercial or political reasons. Current thinking with Apple, Google and Microsoft seems to be: lets come up with our own language rather than conforming to an open standard. And history has shown that companies are quit willing to kill the competition at the expense of developers.

A2 (sequel). And then there's the matter JS interoperability. The investments done in JS library development are so huge that developers will have to live with it, whether they like it or not. Even if tomorrow an ideal CPython-for-the-browser would spring into life, abstinence from the rich dish of good-looking JS libraries would put developers in the position of a principally vegetarian wolf in a sheep den.

Secondly about your mention of the 'dir' method. In general I try to leave things out until I get noticed that people miss certain functionality. When that happens I consider having it in. Often there are several flavors of a method, depending on the parameters. I then try to look if there's a flavor that gives    90% of the functionality at 10% of the code. Allowing dir to sum up all locals would require quite some effort and code. Allowing it to sum up the methods of an object in a way that functionally approaches the way Python does it in most cases, would be a modest effort with a reasonable return. So I'll add a simple implementation of 'dir (<anObject>)' that should cover most use cases.

About your remarks with regard to GIT and version control. It is not without reason that Transcrypt has only a main branch rather than a development branch, feature branches, hotfix branches etc.. In general I've noticed for a long time that I can invest endless energy in tools like this, at limited payoff. So no need to apologize if you don't use GIT optimally, I don't either. If you make something valuable (like the Riot example) for Transcrypt I'll gladly fit it in manually. I for one like the degree of control that this allows.

One thing I don't understand is that I merged your previous commits, but you're still not listed as a contributor. I'd like that to be the case, since in is in the interest of perceived continuity and mindshare the project to become clear that there are multiple persons working on it. If you know how I could solve this, please give me a hint.

In general your contributions are truly appreciated. Sometimes it may take a while before I can evaluate them and put them in Transcrypt.

Contributions to the core I'll always look into very carefully to find out what consequences they might have and I tend to be conservative to include them. With a library  it's simpler. If it wasn't already in and it is accompanied by a relevant testlet, I am very inclined to include it, even if its functionality may not be complete. The fact that someone contributes something indicates that the facility in question is probably used in practice. Since libraries are only included if they're imported, there are no consequences for compactness of code.

Demo's like yours, abundantly commented, clear naming, readme present, are always very welcome.

What would also help acceptance is expanding the gallery. So should you develop something serious (or not so serious) in Transcrypt that's publicly accessible, I'll gladly add a screenshot and link.

Kind regards
J

AXGKl:
Hi,
I share your reasoning about the place of Transcrypt in say, 3, 5 years. I mean I like that I can overload __setattr__ in PyPyJs and I'm sure it will work and I can't, currently, in Transcrypt. But TS (better abbr.) is lightweight enough so that creating a new pragma for the ast walker to emit the js I need is not out of reach within a short amount of time and I know then whats going on and its actually not much, which is good. 
----
Have no idea about how to add contributors, I'm not using github for own stuff, except for a not really serious weekend hack I never did open source stuff, since my baby (as you named it in the podcast) is my company.  Speaking of which, I try to improve our efficiency and code supportability in the (highly customized) projects we do by proposing new tools to the guys - so I'll prepare a little presentation about Transcrypt, will send you a link.
Proposing but not imposing, since I want my guys not using tools because I say so (I could be wrong and quite a few a smarter than myself) but because they want it. And regarding GUI it is still a mess - as soon as a big enough customer really wants customization ( in the meaning of: pays enough for it), then we tend to end up with an unsupportable fork of 'ks of lines of js... a nightmare. Even with component frameworks like Riot, or before, frameworks like angular... Maybe we are doing it wrong since we all think Python primarily. Or maybe its because of the low level nature of JS, reusability and stackability of libs is hard.

And here exactly Transcrypt, especially with its super well designed multinheritance features can finally come to rescue and thats where I'm aiming at. 
----
Therefore: Absolutely in line and happy that you don't just pull in stuff into the core, we would be dependent on it rock solid, deploying Transcrypt at our (Telco) customers. Keep a benevolent dictatorship and consider PRs, like the one I did for dir() (btw: it returns also the methods, I would be happy with it) not as *requests* but rather as clear descriptions of the nature of a specific *suggestion*.

----
Speaking of benevolent dictatorship... on issue, which is not compatible with what you said in the podcast, i.e. closely following Python's evolution.
Let me add an explanation before I get to the point:
I think b.d. *is* the best governance model out there, way better than, say a democratic approach. Only counterexample I know of, very unfortunately, is Python itself - where the father after 20 years of success turned away from the baby, so to speak, guess unaware about the implications of long term industrial use, incl support contracts and legals and 100ks of unmigratable project code out there... Also, for doing I/O intense products there is not even too much technical enthusiasm of migrating to a language, where the basic type is simply wrong, enumerated symbols for humans can't be it. Enough, you get the message. Not relevant for Transcrypt anyway, only relevant for the server side - but I think about doing sth "isomorphic",  where a client component can directly interact with its server counterpart, sharing state.
Getting to the point now, having tried to justify why the server will run in Py2: I like the idea of in process "transpilation", so I adapted the compiler a little bit to run the demos in Python2, the syntax tree is still pretty much compatible. And I can reuse core and libs generated with Py3 anyways (did I mention how much I like your design ;-) ).

So, now finally the question: *Should* you have started the project with Python2, I dare to ask if you could send me that version of the compiler, so that I can compare with mine. 

Cheers,
Gunther

JdeH:
Hi,

I am afraid there never was a Python 2 version. I've been using Python 2.7 for a long time and in fact I still do, since some libs I need, e.g. for medical research are not yet ported. For Transcrypt I planned to follow Python versions, 3.5, 3.6, ...., the 3.5 is just the first version. As soon as 3.6 arives I'll use it's parser, but the differences should be minor, and hopefully backward compatible, I think they won't run the same risks twice.

The split between 2.x and 3.x is troublesome and probably has costed Python a lot of followers. But on the other hand, if you never backtrack eventually  you end up with a hodgepodge like JavaScript. Maybe there would have been a road forward that wouldn't have broken compatibility.

You tell me you let your developers have the last say in which tools they use. I think that's very wise. After all you hire competent people, so it's important to draw their judgement into the account.

I don't think "thinking Python primarily" is wrong. I've programmed long enough to know that not every strong trend is a good trend. For all JS projects I did, I tried to impose a decent, maintainable structure. Looking back at code I wrote several years ago I must say I feel the result wasn't satisfactory, the code is hard to comprehend after some time, no matter how neatly I tried to set it up cleanly. The lack of a decent class system, of multiple inheritance and of a consistent, be it dynamic, typing system was the main cause of that.

As for multiple inheritance, I am glad you for one see the benefits of it. I can't comprehend that a complete language (Java) sprung up on the basis of the idea that interfaces are a superior way to go, especially with respect to polymorphism. Multiple inheritance for me is equivalent with conceptually clear, self-evident design. Of course it can be misused, but that shouldn't be the norm in designing a language.

I think a choice for Python can very well be a choice for common sense. And if it can be mixed with things that others make for JS, why not. I can be productive in it in the way I want, it fits my way of thinking. I am not opposed to JS, it's just not the tool that I can make the best applications with.

KR
J

AXGKl:
same conclusions here ;-)

Regarding the contributions, trying to find the reason why the actual user space tag is missing after the merge, I stumbled over this in Github: Main Page -> Settings -> Collaborators. Its not automatically anybody you pull from. Makes sense maybe and: What an honor if you do it even after finding that I'm obviously not able to switch my editor to tabs... shame on me. 
Note: I can't commit on many contributions, pretty restricted timewise during daytime, and two little ones jumping around me in the early evenings but I think yes, I do look on the contributors list before I invest time into a tool, so it sort of may help Transcrypt, marketingwise.  Thanks.

Will get back to you in the other PR regarding why the sample tag is missing. 

G.

JdeH:
Not any obligation to contribute, don't worry!
I just appreciate the example  and your feedback.
My kids have grown up, spending time with them while they' re young is the best investment.

PR #83, opened 2016-07-16
Master

AXGKl:
fixed the broken template rendering.


AXGKl:
You did everything right with the renaming, I did everything wrong with
1. the dir without the sorting
2. the desperate try to do as much in python as possible, forgetting there are things as bound variables and such.

Details: There is [this](https://github.com/JdeH/Transcrypt/pull/83/files) loop where I create property functions for the riot_tag, thats pretty much the whole point of the riot tag class, besides the lifecycle functions, that these are created as properties like in the python object, not as functions.

In pure Python I would create property functions or use a `@property` decorator but it looks like these are not (yet) in Transcrypt. So I had to do it via javascript - but I *stupidly* thought its smart to run the loop itself in python, forgetting the loop variable (k in my code) was not bound within the getter function in Javascript :-)

And why did it work before? Well...here it gets funny:

my crappy dir() function did not sort. And by pure "luck" the last variable *mine* returned was "label" - so this was returned by ALL(!) getters of the tag (crazy bug) ... and ... label was the only immutable used in the template:

```
<h1>Riot Transcrypt Tag Instance {label}</h1>
```

but AFTER your fix of dir() yours returned 'template' itself as the last one - and so... we saw a template rendered with the template itself as string, inside :-) 

----

Btw: 
Are you ok with me adding an issue for property functions or even property decorators? They are quite a cool feature of Python and looking at core js it seems to me that there could be a rather elegant way of adding them without bloat. I'm asking because this can't be done in a library and I think I rather leave my fingers off the core after that epic fail with dir ;-)

AXGKl:
@JdeH: Totally unrelated, just a hint: You latest commit,  a commit message [made it into](https://github.com/JdeH/Transcrypt/commit/8be41619c1e5b50d2f5273236a5bf22ebfadbb4d) the README.rst ('Bug in hello.py with -e 6 fixed"), thats definitely not intendet by you as feature nr 1 on github ;-)

JdeH:
Hi

Property functions are in already. From the 'properties' testlet:

	class A:
		p = 1234
		def getX (self):
			return self._x

		def setX (self, value):
			self._x = value
				
		def getY (self):
			return self._y

		def setY (self, value):
			self._y = 1000 + value	# Weird but should be possible
			
		def getY2 (self):
			return self._y

		def setY2 (self, value):
			self._y = value
			
		def getT	(self):
			return self._t

		def setT (self, value):
			self._t = value
			
		def getU (self):
			return self._u + 10000

		def setU (self, value):
			self._u = value - 5000
				
		x, y, y2 = property (getX, setX), property (getY, setY), property (getY2, setY2)
		t = property (getT, setT)
		u = property (getU, setU)
		
	A.q = 5678

	class B:
		def getZ (self):
			return self.z_
		
		def setZ (self, value):
			self.z_ = value
			
		z = property (getZ, setZ)
		
	class C:
		def __init__ (self):
			self.offset = 1234

		def getW (self):
			return self.w_ + self.offset
			
		def setW (self, value):
			self.w_ = value - self.offset
			
		w = property (getW, setW)
		
	def run (autoTester):
		a1 = A ()
		a2 = A ()

		a1.y2 = 1000
		a2.y2 = 2000
		
		a1.x = 5
		a1.y = 6
		
		a2.x = 7
		a2.y = 8

		a1.t = 77
		a1.u = 88
			
		autoTester.check (a1.x, a1.y, a1.y2)
		autoTester.check (a2.x, a2.y, a2.y2)
		autoTester.check (a1.p, a2.p, a1.q, a2.q)
		
		autoTester.check (a1.t, a1.u)
		
		b = B ()
		c = C ()
		
		b.z = 100100
		c.z = 200200
		c.w = 300300
		
		autoTester.check (a1.x, b.z, c.z, c.w)
		
		c.w = 400400
		c.z = 500500
		b.z = 600600
		
		autoTester.check (a1.x, b.z, c.z, c.w)

I didn't bother about decorators because I  they're just syntax.

JdeH:
About the commit message:

OOPS... thanx

PR #84, opened 2016-07-18
Master

AXGKl:
Hi, can u have a look at this PR, finding and removing docstrings?

Guess not only for me its somewhat annoying when I can't use docstrings...
I did it within `compiler.py` itself, not in a `import tokenize` prepropessor (as suggested in [SO](http://stackoverflow.com/questions/1769332/script-to-remove-python-comments-docstrings), to not screw line numbers, hope its ok (?)


AXGKl:
added also a builtin plus test for round (require that for the autotest of a time module, which fails without round, would you expect? ;-) )

Thought the test for the builtins make some sense to reside in a dedicated test mod, no?

AXGKl:
@JdeH 
Question also: Any problem adding these:

```
*__javascript__
*autotest.html
*autotest.min.html
```
to gitignore? Otherwise, after running stuff the git status is always hard to read. I mean: IF somebody writes __javascript__ modules then he would add it anyway to git manually...
If you agree we could then remove the 2

```
transcrypt/modules/org/transcrypt/__javascript__/__base__.mod.js
transcrypt/modules/org/transcrypt/__javascript__/__standard__.mod.js
```

JdeH:
Hi Gunter,

About the docstrings: Of course you should be able to use them. Moreover, in addition to  PEP257 I would like to be able to add single- or multiline comment strings anywhere. Even though they're not automatically extracted by any tool, it should be possible to e.g. comment on a complicated formula anywhere in a piece of code by adding a string-like comment, since Python doesn't have 'real' multiline comments like JS. Since separate tools can deal with docstrings from Transcrypt code, Transcrypt doesn't have to do anything at all with any statement that solely consists of a string, so I've picked the following very simple solution, adding only two lines in one place to cover all these cases:

	def emitBody (self, body):
		for stmt in body:
			if isinstance (stmt, ast.Expr) and isinstance (stmt.value, ast.Str)
				continue
			self.visit (stmt)
			self.emit (';\n')

I haven't added it to the  autotest, since this would require manual checking. It would be possible to read back the generated JS in the autotest and check that indeed the docstrings are gone. But checks like that could be done for any generated statement so I left it out completely since it would be overkill. 

About exempting all \__javascript__ directories from version control. I've thought a great deal about that a few months ago. Why I didn't choose that solution is that doing so isn't fail-safe (or inherently safe, if you wish). If I forget to take measures to explicitly include JS-only modules, they'll be thrown away, which could potentially destroy quite some work. I'd like it to be like the braking system of a train: if the pressure is off, the train stops.

So I came up with another solution principle, that's part of the shipment test. Any demo, example, autotest etc. that's part of the Transcrypt distribution will be checked in the shipment test. This solution works as follows:

1. At the start of the shipment test the start time Ts is registered
2. Anything that's part of the TS distro has to be built as part of the shipment test anyhow (or it will eventually fall over unnoticed)
3. So any file written after Ts is generated by Transcrypt, so it shouln't be part of the distro, version control etc.
4. By a recursive treewalk those files, the ones having creation time after Ts, are thrown away.
5. A tough choice was wether or not to include the sfinx docs in this procedure. In favor: Generated docs cause an awful lot of version control garbage. Not in favor: If they're not kept, anyone downloading Transcrypt would first have to generate the docs (or use the on-line docs, of course). To keep things simple I kept the generated docs: Users should be kings, and we should serve them, even if it costs us some. But in the long run the installation procedure could be changed, to generate the docs on the user computer. That, however, would have to be tested with each release under Windows, Linux and OSX, for now adding more complication than I think it's worth.

So, if you're testing a demo that's part of the distro and you don't want to be bothered with all the generated 'garbage':

1. Add your demo to the list in the shipment test:

> test ('demos/nodejs_demo', 'nodejs_demo', False, True, switches + '-p .none ')
> test ('development/automated_tests/hello', 'autotest', True, False, switches)
> test ('development/automated_tests/transcrypt', 'autotest', True, False, switches)
> test ('development/manual_tests/module_random', 'module_random', False, False, switches)
> test ('demos/hello', 'hello', False, False, switches)
> test ('demos/jquery_demo', 'jquery_demo', False, False, switches)
> test ('demos/d3js_demo', 'd3js_demo', False, False, switches)
> test ('demos/ios_app', 'ios_app', False, False, switches)
> test ('demos/react_demo', 'react_demo', False, False, switches)
> **test ('demos/riot_demo', 'riot_demo', False, False, switches)**
> test ('demos/pong', 'pong', False, False, switches)
> test ('demos/turtle_demos', 'star', False, False, switches + '-p .user ')
> test ('demos/turtle_demos', 'snowflake', False, False, switches + '-p .user ')
> test ('demos/turtle_demos', 'mondrian', False, False, switches + '-p .user ')
> test ('demos/turtle_demos', 'mandala', False, False, switches + '-p .user ')
> test ('demos/terminal_demo', 'terminal_demo', False, False, switches)

If you don't want to run the whole autotest (it takes up much time), switch off the lines you don't want by putting a # in front of them.

Use the shipment test with a -c switch, to clean up any generated files afterwards:

python35 shipment_test.py -c

You can of course also run it first without the switch, to be able to inspect the generated files.
After that you run it again with -c to clean up the mess.

JdeH:
Since I'll add ony part of the pull, it's easier to copy it manually than to pull it automatially, so I'll do that rather than merge.

AXGKl:
@JdeH 
hi, thanks for the improvements (regarding round, where I noticed the little error also in tests today)  and the more radical approach regarding the docstrings, which I like as well.

For builtins in general, yes I see the problem that builtins should stay as small as any possible. On the other side its annoying IF you want a function like round and don't get it. Maybe an additional builtins.py would make sense, since you have that cool mechanics of detecting what is actually used and only that is shipped?

---
Regarding the gitignore: Matter of workflow and pretty complex solution you invented but seems to be robust.  I mean, I'm used to explicitly git adding the stuff I did and which should be in, plus even a git reset does not remove untracked files, so _that_ problem of loosing stuff matching an ignore mask was not that present to me => sorry for the suggestion. I can help myself in any case, no problem and understood why you want it in. 

----

Actually those detailed explanations of yours should be tracked somewhere, in a developer docu or sth, more informal, but accessible to others, to precious for a PR discussion. Maybe just directory with named markdown files, copied from here, no build tools, just markdown ?

AXGKl:
> Since I'll add ony part of the pull, it's easier to copy it manually than to pull it automatically, so I'll do that rather than merge.

That was my idea regarding collab on the core: Just regard my Pull Requests as concrete *descriptions* of what I suggest, where you can copy and paste from stuff you deem useful. Then ditch the whole PR, if copy and paste for you is faster than merging.  I'll always pull and reset hard to your subsequent commits on HEAD. 


I just wonder how you'll manage alone, when more guys like me join the bandwagon and send you PRs for all the little details here and there. Downside of having written it so transparently that any intermediate can understand within a few hours only ;-)  Hope you can manage that....

JdeH:
Hi,

**About managing the project:** I've known from the beginning that if this really catches on, cooperation with multiple (maybe many) developers has to be structured somehow. For now I am very glad with all the idea's you and others contribute + solutions or hints of solutions. Having no feedback is like driving a car in the dark without headlights. But yes, there may come a point where some form of organisational / delegational structure becomes necessary, together with a set of clear rules about who contributes to what.

It will be inevitable that I'll divide the time I dedicate to this project between developing, thinking ahead, quality control and managing . Probably there should be a very limited group of people contributing to the compiler, core and builtins and it also should be very explicit what demands are made on the code. So you're right, it is important to lay down past design considerations somewhere. 

As I see it there are at least three stages in a successful open source project.

Stage one is the stage where the originator has to do many things himself, doesn't get much feedback or help etc.. If some people contribute, in this stage it's easy to talk through their contributions with them, test everything, integrate these contributions into the project. It is the stage where the project itself (not only the product) is lean and agile. It is also the stage where the overall direction is set for a long time, by taking the right design decisions with a careful eye on their long term effects. I've tried to draw long-term considerations into account in the design of the kernel, the aliasing mechanism -with regard to additions to (the) JS (namespace) itself- etc.

Stage three is the stage where the project is high profile, many give feedback and many want to contribute. So it's a matter of an appropriate cooperation and selection mechanism, and someone firmly keeping the hand on the wheel with regard to the core and the basic direction of the project. In that stage good inspiration can be drawn from e.g. CPython, Linux, Google Closure.

Stage two I find the hardest. It is the transitional stage. The project starts to take up a lot of time from the originator, still contributions are to coincidental and sparse to impose a structure. There's not yet an "inner circle" of committed developers that are familiar with the guts of the project etc..

At this moment I consider Transcrypt still to be in stage one. But indeed now is probably the time to think about stage two. Whatever stage Transcrypt will get into, for now I plan to see and 'process' all code that goes into the compiler, core and builtins, since that's the only way to keep a grip on quality. It's what I see Linus and Guido doing. Wouldn't want to compare myself with these giants, but I think I should learn from them, even in a very modest way. The fact that they've kept their projects between the white lines for so long has fascinated me for years. In general I find it beautiful to see how serious and committed the open source community CAN be...

KR
J

JdeH:
One word about the compactness of builtins etc. The reason the Google Closure compiler is used, is that it should strip off dead code. Unfortunately currently only the least proficient level of minification currently works properly for all code generated by Transcrypt, even though it generates correct and strickt JavaScript. By the way I don't blame the GC guys for that, TC also isn't perfect.

AXGKl:
yes stage is 2 is the critical one, also because the non technical skills of the project leader will play a decisive factor. I.e. authority. From surveys of successful companies led by the founder still, they found that there is not THE ONE successful style but there IS, always, **A** consistent style plus *authority*. Which some have, many don't. Its obvious that you have it.

Regarding the two giants I can't help commenting, even if I hurt your feelings as a Dutch, that I would like more Linus style: An open 'you are a stupid ...' into the face is way better than being stabbed in the back, with a cheerful smile, like Guido did to the industry depending on the stuff. This will get really bloody when they stop supporting Py2 and there will be two Pythons then.
Linus is famous for his top principle: Never.Ever.Break.Existing.APIs.
Still thankful for Py2 though.

JdeH:
They simply shouldn't stop supporting 2.x but rather integrate it 3.x with a command line switch or even locally using some pragma-like construct, so you can mix the two through an interoperability layer, enabling people to make the transition slowly and gradually if they wish. Shouldn't be too hard. And you should be able to leave either the 2.x or 3.x part out to have a smaller interpreter if you wish.

Given the huge popularity of 2.x this would still be worthwhile. For me one project is enough, though.

PR #86, opened 2016-07-20
time module

AXGKl:
Hi, this would be my shot on a time module for transcrypt. js does offer much but when you are used to pythons' time it might be welcome by people, since js handling is very different...

I tested on my laptop and on a linux server - but much stuff is very locale dependent (e.g. weekday names) and more tests / feedbacks from other areas in the world would be welcome to remove remaining problems.


AXGKl:
Forgot to say: The module depends on the new docstring handling in TC (!).  The  missing split with limit I shimmed, also zfill.

AXGKl:
also, one thing I don't like at all: a central data structure is the struct_time tuple, pretty much everything goes via it. and this contains this 'is daylight saving time' as a value - which is
1. super tedious
2. super expensive to build, involving 2 new Date objects (jan and jun dates of that year for that date in question, in def __get_jan_jun.
I'm not experienced regarding memory management - do you think thats a problem, or are the objects anyway immediately g.c.ed ?

speaking about cases where people build large tables with time values in....

AXGKl:
pls. don't see the small fix:

https://github.com/JdeH/Transcrypt/pull/86/commits/19d823746600e5f2bb1b7e51b7dba315d271809e

JdeH:
Quite some work you did for the time module!

AXGKl:
yeah went  a bit out of bounds especially with these daylight saving infos , that I underestimated, effort wise. and the strptime without empty seperators in the format string.
But 90% compat would be terrible when there is no real hard tech reason why it could not work (like with sleep) and if its a module (which does not bloat the core).  

I wonder if "dual world" constructs like `sleep(sec, callback_function)` would be acceptable in TS - or is the design goal rather: 100% compat to python - if not doable then don't.

Maybe the answer would be sort of a tools module, for such non 100% Python stuff, also for the decorator work Dabo and myself did in https://github.com/JdeH/Transcrypt/issues/92.

from tools import decorate, sleep

?

JdeH:
Hi Gunter

This is a question that requires a somewhat more elaborated answer (that maybe in the near future should be linked to from the distro, in line with what you suggested earlier).

It is clear that both the proposed implementation of sleep and of decorators currently have good use in Transcrypt. On the other hand, decorators probably will find their way into Transcrypt natively before long, and several semantically slightly different implementations of 'sleep' may emerge, let alone the fact that some future JS version might support 'sleep' semantics more closely to Python's (e.g. as part of node.js, I don't see that happen soon in something as event driven as a browser.)

Then there's the question of nomenclature: should 'sleep' be called 'sleep' of should it be called 'pause' to avoid confusion. Should its param be in seconds (in fact I prever SI units everywhere) of should it be in milliseconds (since that would make it an integer on most implementations).

As long as CPython is followed exactly, such choices are simple and unambiguous. This is e.g. the case with the time module that you contributed. It clearly can best be imported simple as 'time', and not e.g. as 'com.axiros.time', since it is upto a high degree a drop-in replacement for CPythons time module, and people simply expect 'import time', based on their experience with CPython.

For the proposed 'sleep' cousin and a good approximation to decorators, however, things are more ambiguous, and I'd like to keep room for several implementations emerging and living next to eachother. I therefor propose to choose one of the following:


Option 1
========
Your proposed 'tools' module could be part of the distribution. In that case it should not prevent anyone else to contribute a module called 'tools' in the future since the name is quite generic. This is why, in addition to Pythons module naming system, Transcrypt supports Java-esque unique, url based naming. In that case it should be imported e.g. as:

'com.axiros.tools' (or in general <anyUrlYouHaveControlOf>.tools. It would then reside in ../transcrypt/modules/com/axiros/tools.

This avoids the name clashes that plague the Python world and would grant you freedom to freely use any module naming system you like, inside the com.axiros or <anyUrlYouHaveControlOf> namespace.

Also anything in the distribution would involve me, at least as a doorkeeper and quality guard. Since I hope the Transcrypt world will eventually grow faster than I can keep up with, while currently acceptable, this may eventually come in the way of fast proliferation.


Option 2
========
Your proposed 'tools' module is not part of the distribution, but rather pip-installable (as is the case with Numscrypt). In that case name clashes are avoided by PyPi, since you would have to find a unique name for your module on pypi.python.org. I've taken a look, and 'tools' is already occupied, but I guess a name like 'transtools' or 'toolscrypt' would still be free.

I can think of many arguments in favor of both option 1 and 2. None of them would be conclusive, however. Still I have a slight preference for option 2, for a tactical reason. Having another PyPi package with the keyword 'Transcrypt' would make it very explicit that it is possible for anyone to contribute to Transcrypt in a totally free way, not depending on the willingness of anyone to include it in the distro.

Ideally such module would eventually would include autonomous testing facilities, e.g. an autotest and a shipment test.


So...

If you come up with a good tools module, I'll gladly include it under com.axiros.tools or <anyUrlYouHaveControlOf>.tools.
But I'd slightly prefer it in this case if you make it available via PyPi, e.g. as 'transtools', or 'toolscrypt' or any name you want.
In that case I'll place a link somewhere in the docs or on the site.

PS. I can imagine the process of uploading anything to PyPi is somewhat of a barrier the first time. The links below helped me:

http://peterdowns.com/posts/first-time-with-pypi.html
https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/

Kind regards
Jacques

JdeH:
BTW The time module is now part of the distro

AXGKl:
1. ty for adding time. I hope it does not break tests in other regions of the world though, there are quite some locale dependent features and I tested only on two systems, changing the system locale though.  

2. I find the idea of transcrypt modules on pip really great. I recently uploaded my [weekend hack](https://github.com/axiros/terminal_markdown_viewer) to pip and it was not so hard. I think that would help proliferation of transcrypt  - and gives you the opportunity to cherry pick from time to time well accepted stuff from there into the distro, instead of Option1 where you'd have to do it "syncronously" and all the checking would be on you. 

But I would not know, currently, how I would parametrize setup tools to create a pip which installs within e.g. com.axiros.tools within the Transcrypt path (and triggers the tests on demand). Here maybe a little chapter in the docu would help? I think when I'm finished with my docu stuff I'll try to write such a page, with my tools module as an example. Will take some days though, still a lot to do before :-)

JdeH:
Hi,

Sorry I have maybe not been fully clear. No need to put your module into com.axiros.tools if you make it pip installable. Just give it a unique name like transtools, toolscrypt, tstools or whatever, and put 'transcrypt' into the keywords list to make it findable.

Both pip and pypi unfortunately do not offer a possibility to use universal url's (that's why I have to insert manually into the distro).

As soon as anyone creates a 'toplevel' module name 'com', noone else can use that prefix anymore.
Of course you  could bypass python's module naming altogether and use something like com_axiros_tools or com_axiros.tools, but for me this feels artificicial.

I think it's best to indeed simply conform to what Pythoneers are used to in this case, and make your module pip-installable using a short, unique name rather than a URL based one.

If ever pypi and pip start supporting URL based names this distinction becomes less important. But I don't expect that to happen. Python is Python and Java is Java and never the twain shall meet... I don't think it has to be so, but unfortunately it is.

AXGKl:
hi, just had a little talk with Stephan, our Python authority, about the subject - I just forward his opinion for you as additional info, maybe there is new information for you.

He suggested to use 'namespace packages'. 

Works like [this](http://www.siafoo.net/article/77#multiple-distributions-one-virtual-package)

So I write a module for transcrypt and say in setup.py `namespace_pacakges = ['transcrypt']`

plus in `__init__.py`: 

```
import pkg_resources
pkg_resources.declare_namespace(__name__)
```
done.

---


This is then understood by all tools like pip, apt-get, buildout and so forth.

E.g. pip will install the package into the transcrypt folder, buildout would include it into the sys.path.

---


Example: All the good old zope packages are all findable in pip under zope.xyz, e.g. zope.event, zope.component, zope.app.publisher (100 more).

https://pypi.python.org/pypi/zope.event
```
2.root@m01:~# pip install zope.event
Downloading/unpacking zope.event
  Downloading zope.event-4.2.0-py2-none-any.whl
Installing collected packages: zope.event
Successfully installed zope.event
Cleaning up...
2.root@m01:~# python
Python 2.7.9 (default, Mar  1 2015, 12:57:24) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import zope
>>> from zope import event
>>> print event
<module 'zope.event' from '/usr/local/lib/python2.7/dist-packages/zope/event/__init__.pyc'>
```

Advantage here would be that all packages under pip would be called transcrypt.xyz.


I would find this a bit better than me picking transtools and the next guy tstools (...)


Cheers,
Gunther

PR #88, opened 2016-07-21
Dictops

AXGKl:
Hi, this one fixes and tests setdefault on dicts plus provides dictionary pop ops, they are superconvenient especially when not using kwargs (args.pop('name', 'defaultname') )

I do the function within py_pop, since the compiler creates a py_pop method on dicts, not a pop, like it was an array. I'm happy with that.


AXGKl:
Hmm, I see that all the other, time module commits are also contained in this PR.
I do new branches, to have seperate pull requests for seperate things. Hope its not too much headache ? As far as I know you it should be no problem, you look at the stuff and copy over to your master branch what you find useful, you don't merge when its about core stuff.

so in this one, just look at the last commit.

But next time i do a git reset --hard to the master branch before i add stuff & send you a PR.

JdeH:
Thanks!
Pop and fixed setdefault commited and uploaded, testcase added also for fix.

AXGKl:
Hi Jacques,

possible for you to also include the update?

https://github.com/JdeH/Transcrypt/pull/88/commits/c7fe53ad3926afd137fd805ba17c0f25d8ff325b

AXGKl:
Also, think I sent already a PR or an issue somewhere: This little fix for in, otherwise it makes it crash when its neither dict or array but object:

```
~/GitHub/Transcrypt/transcrypt/modules/org/transcrypt/__javascript__ $ diff __builtin__.mod.js /Users/klessinger/GitHub/GKTranscrypt/transcrypt/modules/org/transcrypt/__javascript__/__builtin__.mod.js 
133c133,135
< 			return container.indexOf (element) > -1;
---
> 			return container.indexOf ?
> 				  container.indexOf (element) > -1
> 				: element in container

```

(btw: actually in my browsers the javasript `in` is working just like pythons)

JdeH:
Hi Gunter

Sorry to have missed the update.

About 'in', seems to work in all cases:

	// In function, used to mimic Python's in operator
	var __in__ = function (element, container) {
		if (type (container) == dict) {
			return container.keys () .indexOf (element) > -1;									// Array
		}
		else {
			return container.indexOf ? container.indexOf (element) > -1 : element in container;	// Array, string or object
		}
	}
	__all__.__in__ = __in__;

Will include in next commit.

JdeH:
There's a fundamental problem with 'in': CPython uses hasattr rather than in! So finding out if something is in an object should use hasattr, also if it's a JS object.

So I reverted the change and left __in__ in as it was.
If you think I am mistaken here, please let me know.

JdeH:
dict.update was committed

AXGKl:
hi, see the argument regarding the in, but let me add a comment into the  https://github.com/JdeH/Transcrypt/issues/96 issue, since its sort of an important thing for others to see as well.

JdeH:
Ok please do

KR
K

PR #157, opened 2016-11-05
removed diff marks

AXGKl:


JdeH:
I'll logout now, otherwise it will keep me awake...
Will try to pull tomorrow.
Have a good weekend!

AXGKl:
so. same thing as before, 4 changesets in one PR. I'm sorry but there was not other way. The good thing: no conflicts you can automerge.

AXGKl:
Regarding git sth:

I think you should gitignore the _build directory of sphinx. Reason: Every little change will result in new binary files created and git can't diff them, so has to add them to the object db in .git directory.  See here, e.g.: https://github.com/QQuick/Transcrypt/commit/1d34a09ff822b22988fc9abc0145c67b054d7378

Currently we are at 76 mb already in the .git directory, but only 20 megs of actual filesystem code base (excluding the 5 megs of the  _build dir) and it will grow massive with any change of sphinx docu text.

=> I know its convenient and as the author you anyway don't care about 100 megs on your harddrive for sth which consumes 50% of your free lifetime.  I also don't care actually. But we are in the start of phase 2 and this is the absolute first impression of people with the code base. 

=> if you want the convenience that you build on your laptop and just push the result to static server which has not sphinx toolchain then I suggest you split documentation off, into a seperate repo. 

If thats not the problem then I'd gitignore _build

JdeH:
The large diffs caused by doc updates have bothered my for  a long time. I didn't want to shift the burden of building the docs onto the user. But since they're available on-line, probably not many people are using the downloaded docs anyhow. So rather than having the (limited) complexity of maintaining another repo,  I think indeed git-ignoring the _build dir is a good option. After all that won't prevent the docs to be part of the pip-installed Transcrypt, if I'm right.

Thanks for pulling me over the border on this point, couldn't make up my mind. _build will be in the git-ignore and the online docs will be leading for people downloading from git. Pippers will have them locally.

JdeH:
**Merge and pull seem to have gone fine and everything works. Looks like a good job to me!**

***

I have made some cosmetic mod's to the classes testlet:

Made some names a bit longer since it took me some time before I understood what r, c etc. meant. Being a bit picky on that, since the testlets also serve as doc.

For the same reason I've put the complicated stuff at the bottom of the testlet (and added labels to the output). Want to prevent beginners thinking classes are something difficult.

***

Augmented function isinstance works fine in the autotest.  Now I want to attempt to make it more compact (if possible, no idea). Not sure it will work correctly for JS types after that, so I'd like get back to you for feedback on having broken something, if that's ok with you.

Maybe we should have a test for it added to the manual (eyeballing) tests, since of course correct handling of native JS types cannot be back to back autotested with CPyton. If it can't be smaller it'll stay as is, since indeed being able to use instanceof on any object is a blessing.

***

Time module works fine on Windows now. Will also test on my problematic linux and maybe add a workaround, since other linux installs may suffer from the same. Think I can do that with almost no code. 

***

I need some explatation on the README.md of the hello autotest with regard to 


> Thats why you have to add the current working directory to have `autotest.py`
> find the testlets *outside* the Python runtime, not within, via `sys.path.insert`.
> 
> On POSIX systems this is the standard way: `export PYTHONPATH=".:$PYTHONPATH"`
> 

"add the current working directory" where exactly? I am missing the connection here.

***

Finally: Since your PR was commited, I'd expect you to be added to the contributers list automatically, but still this is not the case. Some time ago my own commits also weren't visible. The reason for that had something to do with different email addresses.

I forgot the details, and no idea why this must be so complicated, but when I changed my email address known on GitHub to something I'd used earlier in conjuction with Git (sorry for being so vague, but I really have no idea of the magic behind this) all my commits became visible. Maybe a similar trick works for you as well.

KR
J

AXGKl:
Regarding '.' into PYTHONPATH:

```
~/GitHub/AXGKl:Transcrypt/transcrypt/development/automated_tests/hello $ unset PYTHONPATH
~/GitHub/AXGKl:Transcrypt/transcrypt/development/automated_tests/hello $ /Users/klessinger/GitHub/AXGKl\:Transcrypt/transcrypt/run_transcrypt -r autotest.py 
Transcrypt (TM) Python to JavaScript Small Sane Subset Transpiler Version 3.5.235
Copyright (C) Geatec Engineering. License: Apache 2.0

Error: cannot run autotest.py using CPython: No module named 'testlet0'

Aborted
~/GitHub/AXGKl:Transcrypt/transcrypt/development/automated_tests/hello $ export PYTHONPATH=".:$PYTHONPATH"
~/GitHub/AXGKl:Transcrypt/transcrypt/development/automated_tests/hello $ /Users/klessinger/GitHub/AXGKl\:Transcrypt/transcrypt/run_transcrypt -r autotest.py 
Transcrypt (TM) Python to JavaScript Small Sane Subset Transpiler Version 3.5.235
Copyright (C) Geatec Engineering. License: Apache 2.0


Ready
```
-> testlet not found if current workdir is not the PYTHONPATH.


----

regarding collab: did never change my email addy :-/

just tried on my little weekendhack repo and it seems you can invite me:

screen shot 2016-11-06 at 18 34 25

In the perms you can say: read only.

-> click on settings of Transcrypt Repo and there should be this collaborations and teams menu then.
----

Regarding making isinstance more compact, just some thought: 

I WANTED to say actually: Don't waste time here, since you could completely omit the function as well if you see no usage of it in compiler js, i.e. add another indirection for builtins.js, sort of a template system with functions just added if being used actually, there are different users with different programming styles. 

But then came to my mind that such an approach would possibly prevent an evolution of Transcrypt towards a 'market place' of precompiled modules, I mean downloadable from the browser directly. 

That would mean then that builtins in deed would have to be ripped appart into modules. ..

Just a thought which you could keep in mind we are far far away from problems with size of builtins, the ruby transcrypt e.g. is at 500 kb....and pyjs you already commented elsewhere about size ;-)

PR #161, opened 2016-11-07
Master

AXGKl:
Continuous Integration for Transcrypt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+dist: trusty
+language: java
+before_install:
+    - sudo apt-get install python3
+    - sudo apt-get install nodejs
+
+before_script:
+  - "export DISPLAY=:99.0"
+  - "./transcrypt/development/continuous_integration/run prepare"
+
+script: "./transcrypt/development/continuous_integration/run set set1"
+
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,5 @@
+![](https://api.travis-ci.org/AXGKl/Transcrypt.svg?branch=travis_test)
+
 Transcrypt is a tool to precompile a fairly extensive subset of Python into compact, readable Javascript. It has the following characteristics:
 
 - Allows for classical OO programming with *multiple inheritance* using pure Python syntax, parsed by CPython's native parser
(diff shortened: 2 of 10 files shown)


AXGKl:
Hi, sorry for the short commit notes, had to microcommit a lot to test what I did:

Let me explain. 

1. What did I do (Risk):
If you merge this then nothing happens with transcrypt, just one new directory (development/continuous_integration) and a file at toplevel, `.travis.yml`.

-> should be no risk what so ever.

2. Feature:

You as the owner can add Transcrypt repo now at travis.io. 
Pls do that, its free and costs you 10 minutes, you don't have to configure anything, I did it (in .travis.yml).

What we get is the possibility to have our automated tests run
- at each push to Transcrypt (or my or someone elses clone). Any branch is tested.
- the tests are real: They install java, python, node plus
- a headless X server which allows you to run firefox

The testscript also starts a real app server, so that later server modules can be tested as well (and it can guide firefox through the test sequence w/o the need to restart all the time)

I hope and think that this will help users in the future with their Transcrypt forks when they play around. The shipment test is nice and manaual tests will never be fully replacable but I think with an automated firefox and some server sided javascript we can automate really the button clicks on demo guis with this. 

Btw: travis delivers than this little icon which others put on top of their README.md which tells everybody if hte current status in the Repo builds or not....


=> I really really think CI is a must have for phase2



---


check it out: https://travis-ci.org/AXGKl/Transcrypt/builds

screen shot 2016-11-07 at 10 29 08


Yes right now only 3 tests, hello, time and dicts, but will add more soon

AXGKl:
I'm awake since yesterday morning ... wanted to learn that also for my stuff, you know when you have an app server controllable on the other side PLUS the client and a full blown browser which you control via the app server by adding 2 lines of js into autotest.js...

Looking forward to get more tests in and assertions for the demos so that they are also auto...

JdeH:
I've created an account and hooked it to Transcrypt.

This is what I get at a commit:

travis

Have I done what you need to be able to continue?
Or do I miss the point completely and do I have to do something very different?

PR #163, opened 2016-11-07
better readme

AXGKl:
sequence diag was missing

### DIFF
--- a/transcrypt/development/continuous_integration/README.md
+++ b/transcrypt/development/continuous_integration/README.md
@@ -58,6 +58,58 @@ The pidfile of this wgetter is killed when there was success.
 
 
 
+## Sequence
+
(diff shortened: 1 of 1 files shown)


AXGKl:
sorry I begin to bug you.  You merge so fast today and I'm slowing down... I'm off for a few hours now...

JdeH:
You don't  bug me, its fun cooperating and you're being very productive

AXGKl:
hi, did you WANT to change the permissions of continuous_integration/run and test_server.py into non executable?

screen shot 2016-11-07 at 17 35 53

and from there: 


screen shot 2016-11-07 at 17 36 20


Or was this an accident / some windows related problem ?

AXGKl:
effect is:

screen shot 2016-11-07 at 17 38 04


would be good if we could keep those 2 executable. I could fix it via .travis.yaml but I'm using the tester also elsehwere and then I also have to fix there after updates ....

AXGKl:
perfect. 

I see now yours and my builds in my forked repo. cool.

btw: You could add a badge to your website now or later when we have more tests:

```
.. image:: https://travis-ci.org/QQuick/Transcrypt.svg?branch=master
    :target: https://travis-ci.org/QQuick/Transcrypt

```

JdeH:
Executable? I didn't intend to change anything. I suspect Windows...

JdeH:
Hi,

While I think CI is certainly handy, I  am not so fond of the common use of these badges.

- If they're green I don't believe that there's always a solid test procedure behind them.  So they make me suspicious:  What exactly is OK?
- I see a lot of projects where they are red. Made me think something very wrong with them, but it may only be an intermediate commit.

AXGKl:
ok, lets omit the badge. In the state of the tests right now ( 3 tests only) it would be really only a 'marketing' measure, maybe like the slogan.

JdeH:
Not all tests are easily remoted or even automated. Nevertheless the ability to monitor test status seems to me an important part of CI and on-line cooperation. For now I'll let it all sink for some time. I can tell you I take the possibilities that this may offer very seriously.

PR #164, opened 2016-11-07
Master

AXGKl:
Reverting the executable flags.


AXGKl:
hi, did follow this to nail the exe flag into the git db: 

http://stackoverflow.com/a/13593391/4583360


Wow, no I see also run_transcrypt is not executable anymore.... will do the same, moment...

PR #165, opened 2016-11-07
Master

AXGKl:
run_transcrypt also +x again


AXGKl:
I really really hope this remains after your next push from windows. 

otherwise you might need to do this on your machine: 

http://stackoverflow.com/a/6476550/4583360

JdeH:
Git hopefully tamed...

`git config core.filemode false`

Have made a push so you can test it

PR #166, opened 2016-11-07
Master

AXGKl:
hi, added all non es6 auto tests and made the frontend a bit better.


AXGKl:
the exceptions test breaks and I can reproduce:

```
~ $ cat r.py 
assert 2 * 8 / 4 == 2, 'asdf'
```
does not produce javascript

```
	(function () {
	}) ();

```
switches: `-bn -c`
-> test fails. Any Idea ?

AXGKl:
ah - sorry for the early PR, I must *first* push to my fork then see, then PR.

AXGKl:
Now it builds, sorry again, made PR before I pushed to my own fork.

Do you see the same as me here: https://travis-ci.org/QQuick/Transcrypt

which is:

screen shot 2016-11-07 at 22 26 31

-> you can merge now it builds.

AXGKl:
ok, working. I'm really not green but I also really don't want to know how many watts one single push to github now burns :-/

AXGKl:
exceptions test excluded, since in my view there is a bug (?)   

The test list I duplicated into here: https://github.com/AXGKl/Transcrypt/blob/master/transcrypt/development/continuous_integration/set1

JdeH:
Have merged pull without check, since all files inside or about continuous_integration.
No reason to keep you waiting, will take a good look tomorrow!

AXGKl:
dankewell
the next days I have to focus on a few more boring stuffs, have an important presentation to do on thursday... => might be off a few days, lets see.

JdeH:
Good luck with the presentation

JdeH:
Exception test: it requires the -da switch (debug assert), since it also tests assertions, which are off by default.

JdeH:
Tried the -da switch, should have solved the problem but didn't. No idea why.
Have tried to obtain or view Travis log files but this seems to require AWS S3 storage.
So gave up for now.

PR #167, opened 2016-11-08
-da flag for compiles

AXGKl:
hi, -da is in, all works. See diff.

### DIFF
--- a/transcrypt/development/continuous_integration/set1
+++ b/transcrypt/development/continuous_integration/set1
@@ -14,7 +14,7 @@ transcrypt/dictionaries
 transcrypt/div_fixes
 transcrypt/div_pulls
 transcrypt/docstrings
-#transcrypt/exceptions (assert 2 * 8 / 4 == 2 fails, nothing compiled)
+transcrypt/exceptions
 transcrypt/extended_slices
 transcrypt/general_functions
 transcrypt/indices_and_slices
--- a/transcrypt/development/continuous_integration/test_server.py
+++ b/transcrypt/development/continuous_integration/test_server.py
@@ -180,7 +180,7 @@ def run_test(filepath):
         with open(fn + '.py', 'w') as fd:
             fd.write(ci_at % (test, test))
 
-    for flags in ('-r', '-bn -c'):
+    for flags in ('-r', '-bn -c -da'):
         run_t(flags, './%s.py' % fn)
         #env['PYTHONPATH'] = '/root/Transcrypt/transcrypt/modules:.'
 


AXGKl:
ups. conflicts... I check

AXGKl:
again a too early PR, travis breaks with the fix as well :-((   

The strange thing: On a local test -da did work !!!

Can only be because of a browser difference, I check with firefox (in travis they have firefox 40 + something)

AXGKl:
btw: Local testing on the laptop is really simple (at least on linux, did not try win, have no win):

./test_server 8080    starts the server
then:
http://127.0.0.1:8080/do/spec1   or /chk/time  or /chk/.exceptions

AXGKl:
ok, local test sort of confirms travis:
FF mac osx 49.0.2 works

but FF ESR (Enterprise Stable Release) mac osx 45.4.0 does not:

screen shot 2016-11-08 at 18 34 29


guess this can't be fixed and we must keep the test excluded (?)

JdeH:
I changed my mind and added the CI icon to the GitHub frontpage, together with some explanation about what it currently does and doesn't mean, since they can already be of value to contributers.

I've tried to find out how we can get hold of Travis'es 'Artifacts' (what a word...) without going via a AWS S3 store. If we can get Travis to e.g. mail back results to us, or store them in GitHub, the CI tests become much more valuable. But I couldn't find info on where Travis exactly stores the Artifacts.

AXGKl:
yeah, artists are not bound to what they said yesterday but only to what they feel is right (saying this to my CEO every second day ;-) )

regarind the artifacts: why? isn't it way simpler to just run the test with a specific git version locally ?

AXGKl:
back in 30.min, have to buy some food for the evening, relocated into the office for a few days ;-)

JdeH:
Maybe it can be fixed, now that we have some feedback, I'll look into it.
I do not understand the role of a browser in the CI test.
Seems I have missed something.
Are you loading the compilation results actually into a browser?
And is the error not a compilation error but an error reported by the browser trying to run the JS?
Seems you're even further with this then I thought...

JdeH:
Probably you'll need all available time for you presentation...

With regard to local tests

- Yes it's simpler, but the results are not easily available to anyone committing. That's what I like about Travis.
- Even when testing locally, I'd like to automate the tests even more than they already are. That has nothing to do with CI, but everything with saving time spent on testing. But as a byproduct a larger part of the tests may be able to conclude with a simple yes or no, only saving a screendump if something went wrong.

This is all just a long term fantasy, but anything we automate will pay itself back. In the beginning I thought myself to be quite clever with the back to back autotest. What I didn't realize at that point is that they autotests were an absolute necessity. They've literally caught hundreds of bugs without much effort. On one hand testing is a dull subject, and even as a quality manager (a job I did for one of my customers for 10 years) I have always underestimated its crucial role.

I am not saying CI/Travis is the best solution to the testing problem. Maybe we should be satisfied with 'it runs' for dayly builds. But with the company I worked for they went a lot further. Everyone was automatically spammed with screendumps of problems that his code caused in the dayly builds, als semantically (computational output.)

AXGKl:
hi, back. I seem to have not made clear enough that the whole cI testing is 100% your autotest framework. Just automated:

- It starts a real firefox browser in an X-server (!) 
- which hits an umbrella URL on a little server also started in the test 
- whose only job is to guide the browser to the different autotest.html files given in set1 file, and to compile the autotest.py files, clear.
- plus it inserts one litte line of javascript after your generated js which checks if the browser displays the 'succeeded' in the message div.

=> So all your work for autotesting and generating those html files is absolutely the basis for everything. They are run by a real browser.

and you can run them also using your real browser, just hit the app server at /do/set1

AXGKl:
and yes the screendumping is also possible, this is done by a service called selenium, would be also available in travis. [here](https://docs.travis-ci.com/user/gui-and-headless-browsers/) more about this.

But I thought by just appending some tester javascript to the generated ones checking if all is ok on the page we are good to go for now. For sure we could even click some buttons and stuff like that...

JdeH:
I am VERY impressed!!!
Deutsche "grundligkeit" at its best. :)
I can make the explanation about the icon somewhat less careful.
Can hardly believe what you achieved literally overnight...

BTW Tried to solve the exceptions test problem.
Curious about the CI result now.

AXGKl:
looks good, ESR working nice.

deutsche gruendlichkeit? naaa...I'm not the guy for that

the code in the test_server is ugly, most of the time I spent on coming up with that *method* of running through the tests. Most time I burned for nothing, trying to do it with phantomjs, i.e. just a simulated browser. Dependency problems. Then firefox hitting the  file:/// ..../autotest.html like you do in shipment and multiple opens on travis I did not like, regarding reproducability (timings and such) . And serial hits  took way to long to start it up all the time.
Plus this problem of not "knowing" what firefox sees, was even trying to export the html into a pdf file, some people do that. 
Only then I came up with that app server approach, where the transcrypt js is added by the checker line and the redirection code and the app server checks the result that way via a get from the js.

trying to say: I'm not IT, not gruendlich and not nearly as smart like you - but I can stand a lot of pain, gets you also a long way :-)

JdeH:
Don't think it all comes for free with me...
Some parts of Transcrypt were just a matter of trying over and over again, investing countless hours.
E.g. the metaclasses didn't come in a lucid dream, but by first doing it the wrong way four times.

But anyhow, I think what you achieved now has a big impact on the project.
It is another step from a one person project to a true open source project.
I am fascinated by the quality of the feedback from some people.
All these minds in parallel should be able to achieve something...

BTW I seem to have choked Traivs by replacing all the distinct testlets  by one reference to transcrypt. Did that because I assumed that's how hello and time worked. Seems high time I study your continuous_ingegration code a bit more thorough.

I'll try to revert set1 to its proper state for now...

JdeH:
Seems to be working again...
Time for a cup of coffee.

AXGKl:
uh oh and then me = saying metaclass should be straight forward... I beg a pardon.

the full transcrypt module test I did split into seperate ones since it did fail. Also right now here:

screen shot 2016-11-08 at 20 17 12


If you can fix that then you could include it like you did but you either have to delete the single ones or comment out by prepending #  - the ''' style exclusion it does not understand, just really stupidly simple hacked together:

```python
 88             tests = [k.strip() for k in tests.splitlines() \                                               
 89                      if k and not k.startswith('#')]               
```

JdeH:
A, the accuracy devil again, mother of all autotest plagues...

AXGKl:
we could be doing the compile switches tests more declaratively, like:

```
transcrypt/dictionaries, -e 5, -e 5 -da, -e 6
```
and just gather the switches in `145 def run_test(filepath):`, and run transcrypt once per switch given in the test spec file.

Or do it only if the transcrypt overall module test fails, to know what failed directly. anway, not today.

AXGKl:
but REALLY cool actually would be to know WHICH test assertion failed. If you could deliver that in another div.. or, check for colorcode red in the DOM and read to next comma ... hmmm. anyway, not today.

PR #168, opened 2016-11-08
Master

AXGKl:
hi, just keeping the single tests available (not for travis which runs set1 but for local tests with the browser). 

funny: the full test is still having that math issue on my laptop but not in travis. and only when its included in the full test, when run standalone module_math it is working.


AXGKl:
argh, 74 changed files wtf ... I love github - but I hate git. forget this PR, I make a new fork.

PR #175, opened 2016-11-16
Master

AXGKl:
Hi, this fixes a crazy errror reported here:

https://github.com/QQuick/Transcrypt/issues/173


AXGKl:
@JdeH : Please keep in mind if possible that the CI tester checks the message div like this:

`if not 'green' in res and not 'succeeded' in res:`

=> so if you would remove green AND change "suceeded" all tests of CI tool will break. 

Sorry for the screenscraping approach but as long as you don't change both of these it should be working.

JdeH:
I'll be sure to keep those 'keywords' in the message id.
Maybe later on we do something with an invisible element, but there's absolutely no hurry with that (if at all).

Issue #171, opened 2016-11-14
XVFB dependency & issues for dev


RCL-Carl:
Hi, 

I'm new to this project so maybe this is a repeat but I didn't see it in the issues list, so hear goes nothing. 

# Concern Num 1
I'm trying to get a dev environment setup so that I can run the automated test suite. After looking at the .travis.yml file, there are some dependency requirements but I found this doesn't capture everything. On my debian machine I had to install xvfb in order to get this to run: 

`$> sudo apt-get install xvfb`

Once I got this installed, I could run the 'run set1'

# Concern Num 2
'run prepare' attempts to use sudo. In Travis, I'm sure that works find but as a developer, when a script asks me for sudo - that concerns me because any subsequent calls to sudo will run without me knowing about it. I know I can pull up the script and look at it to figure out what it is but that isn't necessary in this case. All the script should have to do is check if the symlinks exists, something like:

```
    if [ ! -e /usr/bin/python3.5 ]
    then
        sudo ln -s /usr/bin/python3 /usr/bin/python3.5 2>/dev/null
    fi

    if [ ! -e /usr/bin/nodejs ]
    then
        sudo ln -s /usr/bin/node /usr/bin/nodejs 2>/dev/null
    fi
```

# Concern Num 3 

The 'run set1' command launches the python test server and xvfb command for the headless firefox which is cool, but the script does not cleanup after itself when it runs. Again, I get it - when this runs in travis/docker, it is going to clean up after itself by just bringing down the instance. Unfortunately, this means to run these tests I have to run docker or use travis cli. I can run the 'run set1' test script at the command line in a normal debian box. The results are as expected but now because the python server and xvfb commands are still running, I have to go in and kill them individually before I can rerun. 

I've made some changes to that run script to capture the PID of the python test server and PGID of the xvfb-run command (this script does not kill its children correctly, which sucks, but there is  a way to deal with it using setsid). I can now kill these processes after the test script runs. I'm planning to add a command line argument flag to tell the test script to kill the processes after it has run so that the default behavior does not change. 

# Request

If I submit a pull request for these three items, will you accept it?

RCL-Carl:
>  that concerns me because any subsequent calls to sudo will run without me knowing about it.

I may have spoken too soon on this - I think it asks you every time in the script. But I think checking if the a file or symlink exists for those before blowing them away is still a good idea.

AXGKl:
[...]

PR #177, opened 2016-11-16
Gracefully Kill Children of CI run script

RCL-Carl:
This PR targets Issue #171 

I've added code to capture the PGID of the XVFB process group, PID of the test server, and code that kills these child processes when the script has completed. I've also had a handler for Ctl-C so that if the user attempts to kill the hung process the children get cleaned up properly.

### DIFF
--- a/transcrypt/development/continuous_integration/run
+++ b/transcrypt/development/continuous_integration/run
@@ -55,12 +55,27 @@ export DISPLAY=:99.0
 url="http://127.0.0.1:$port/do/$tests"
 
 echo "Testurl is: $url"
-/usr/bin/xvfb-run -a firefox  "$url" &
+# Launch XVFB detached from this process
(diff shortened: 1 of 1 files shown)


JdeH:
@AXGKI: I assume this is the pull request you talked about:

> checked your PR, from my side you can send it to Jacques, its ok to check for presence of the node and python binaries and only if not present do the symlink with sudo, plus the formatting changes and killing of processes

So I'll merge it rightnow.

If a PR involves CI / CI testing, please confirm **in that PR (so right here in the comments)** that I can merge it, so that I can avoid mistakes, merging the wrong thing. Since you're the initiator of the CI, I'd like some kind of "formal approvement" for that. I would'nt want to break anything. If you approve of a merge in the CI I'll take only a fast, superficial look and commit it a.s.a.p..

Issue #184, opened 2016-11-17
Improved Browser View of AutoTest Results


RCL-Carl:
Hi, 

So I've been working with the autotest browser results and it is very frustrating to compare the results. Additionally, the browser results as shown don't allow an easy way to determine where a particular issue is. In addition, the current browser view only highlights the first error that it encounters. To fix this I've refactored the "autotester" HTML generation to look more like the attached image:

<img width="713" alt="screen shot 2016-11-17 at 9 03 13 am" src="https://cloud.githubusercontent.com/assets/15036736/20399130/b46bf1d2-aca4-11e6-89b3-f70dbf2a51b6.png">

The idea here is to use the [inspect](https://docs.python.org/3/library/inspect.html) library in python and the exception stack trace in javascript to determine the location in the file where an autotester.check method is called. (I have not implemented the python side, but you can see the javascript side in the image). The aesthetics for this are obviously pretty rough still but that could be fixed with a CSS script. 

Additionally, I've refactored the output to use a html TABLE instead of a list of values. I haven't worked out all the bugs in this yet, but this method allows for the results to be compared directly and hopefully, when I have the python inspect stack trace shown in the first column, it will allow the developer to identify exactly where the issue is. 

Ideally, I would like to make the table expandable via a button so that tests that pass completely are shown collapsed and the tests with errors are shown expanded. The user could then click a button to expand/collapse test suite results. 

Would you accept a PR for this feature?

JdeH:
Looks very promising.
Am I right that the testlets stay as simple as they are rightnow?

I can only make a final decision based on the end result.
But for now this sounds like a very useful improvement.
So it stands a high chance of being accepted.

RCL-Carl:
> Am I right that the testlets stay as simple as they are rightnow?

That's correct - nothing in the testlets would change. This is change would only affect the frontend and adds some additional meta data in the `autoTester.check` method for the code location information.

JdeH:
What exactly do you mean by adding metadata in autoTester.check? It would pick up the location where it's called I guess? But syntactically the call would remain as simple as it is now, right?

RCL-Carl:
the call signature in `autoTester.check` stays the same. 

Excerpt of check : 
```
def check(self, *args):
    position = getCallerLocation()
    item = ' '.join ([self.sortedRepr (arg) for arg in args])	
    if __envir__.executor_name == __envir__.transpiler_name:
		self.testBuffer.append ((position,item))
    else:
[...]

PR #188, opened 2016-11-18
Autotestbrowsermod

RCL-Carl:
This PR is intended to address Issue #184 

In the PR, I've also added a small commit to generate an exception traceback in when transcrypt attempts to run the passed source in python. Without this traceback it is quite near impossible to debug what the problem is. 

The major portion of this PR is a commit refactoring the auto test HTML generation. This  adds the new table view concept. See the attached images.
screen shot 2016-11-17 at 3 57 57 pm
screen shot 2016-11-17 at 3 58 15 pm

This PR also adds some code to capture any exceptions that are generated by any of the testlets and it will attempt to show the stacktrace for that exception if it exists. Currently this won't capture exceptions coming from the JS runtime, but it will capture exceptions coming from the python code. This is just base line feature right now. 

The table formatting of the generated output currently limits the table to max body size of the HTML page. This means that for very long strings there can be problems. I've tried to strike a balance but it seems that firefox and chrome behave differently in this regard :-/ 

Currently, the position where the `check` method was called is recorded for both Python and Javascript. Unfortunately, the javascript method location is not accurate. Depending on the call stack, it is very difficult to figure out where in stack the method call is coming from. It works sometimes, doesn't work others. This requires some more finesse. The Python call location however is accurate from what I can tell and I think that is more important for right now.

### DIFF
--- a/transcrypt/__main__.py
+++ b/transcrypt/__main__.py
@@ -111,6 +111,7 @@ def main ():
 					return setExitCode (exitSuccess)
 			except Exception as exception:
 				utils.log (True, 'Error: cannot run {} using CPython: {}\n'.format (utils.commandArgs.source, str (exception) .replace (' (<string>', '') .replace (')', '')))
+				utils.log(True, traceback.format_exc())
 				return setExitCode (exitCannotRunSource)
 		else:
 			try:
--- a/transcrypt/modules/org/transcrypt/autotester/__init__.py
+++ b/transcrypt/modules/org/transcrypt/autotester/__init__.py
@@ -20,16 +20,119 @@ errorColor = 'red'
 highlightColor = 'yellow'
 testletNameColor = 'blue'
 
+def getFileLocation():
+	""" This function needs to crawl up the stack
+	and find out where the grandparent caller of
+	this function was in the source code of either the
+	python or javascript, depending on environment.
+	"""
+	if __envir__.executor_name == __envir__.transpiler_name: # js
+		s = None
+		__pragma__('js', '{}',
+			'''
+			var e = new Error();
+			if ( ! e.stack ) {
+			  console.log("MAJOR ISSUE: Browser Error lacks Stack");
+			} else {
+			  s = e.stack;
+			}
+			''')
+		# Now we will process the stack to find the grandparent
+		# calling function
+		# @note - I'm explicitly not including a 're' module
+		#    dependency here
+		frames = None
+		__pragma__('js', '{}',
+			'''
+			var linereg = new RegExp("\\n\\r|\\n", "g");
+			frames = s.toString().split(linereg);
+			''')
+		if ( frames is None or (len(frames) < 2)):
+			__pragma__('js', '{}', 'console.log("Failed to Split Stack");')
+			return("UNKNOWN:???")
+
+		# @note - if the call stack in transcrypts javascript
+		#   translation changes then this index may need to change
+		#   @todo - need more work here to determine this because
+		#     this is fragile
+		gpFrame = frames[5]
+		# This regex splits the string coming from the javascript
+		# stacktrace so that we can connect the file and line number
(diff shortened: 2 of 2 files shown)


JdeH:
@RCL-Carl 

This really works beautifully!

It makes it much easier to see:
- what is wrong
- where it's wrong
- multiple errors at once

So your PR is gladly accepted.

Playing around with it a little bit I can think of two improvements, probably tiny compared to the amount of work you already put in:

1 It would be handy if, apart from 'Error = <nrOfErrors>', there would be a unique marker at the error locations to search for, rather than scroll until you found it. I often use '!!!' for that, but any marker with a low probablitiy of being present in the 'regular' output will do.

2 It would also be handy if, when optionally there would be a possibility to hide output lines and headers that contain no error. Giving all correct lines and headers with error count 0 a special css class and then having 2 buttons on top of the page to switch those from visible to hidden or something like that would probably do the trick.

I think we should at least have 1 and having 2 is a luxury. Since you're very much into this code now, would it be OK for you if you added point 1 (and maybe point 2, if it's not too much work).

JdeH:
@RCL-Carl BTW With regard to this remark in your code:

```
		# Needed because Transcrypt imports are compile time
		# @note - I really want to differentiate python from
		#   javascript environments - I don't care what version
		#   of python - need to determine what that symbol should
		#   be
```

The following code shows how to make the distinction:

```
from org.transcrypt.stubs.browser import __envir__, __pragma__, __symbols__

if __envir__.executor_name == __envir__.transpiler_name:
	document.getElementById ('output') .innerHTML = 'Using Transcrypt'.format (__symbols__)
	
__pragma__ ('skip')
print ('Using CPython')
__pragma__ ('noskip')
```

This rather asymmetric looking approach requires some explanation.

I've experimented with adding symbol \_\_python\_\_ when running interpreted Python (-r switch) and symbol \_\_transcrypt\_\_ when running Python compiled to JS.

While this is simple to do, it does **not** allow a construction like:

```
__pragma__ ('ifdef', 'python')
doDesktopThings ()
__pragma__ ('elif')
doBrowserThings ()
__pragma__ ('endif')

```

The reason for this is that the CPython interpreter has no idea how to conditionally skip (rather than not execute) code based on  a pragma. To learn it that trick, all modules processed by CPython must be preprocessed by a 'macro preprocessor'.

But, fortunately, CPython doesn't need to skip (rather than not execute) any code. Since everything in CPython happens at runtime (because interpreted rather than compiled), just putting the code under an if is enough.

When only making the distinction at runtime is needed you can use:

```
if __envir__.executor_name = __envir__.transpiler__name:
    doBrowserThings ()
else:
    doDesktopThings
```

Changing this to:

if __envir__.transpiler_symbol in __symbols__:
    doBrowserThings ()
else:
   doDesktopThings ()

wouldn't offer significant (if any) advantage. So I'd like to stick with one way to do this.

I'll also add an `__envir__.interpreter_name` to allow a bit more freedom of expression.

But in case you really want to statically skip a piece of code in Transcrypt, just use `__pragma__ ('skip')`.
And, as said, in CPython you cannot statically skip anything without introducing a laborious preprocessing stage, which I want to avoid.

RCL-Carl:
> 1 It would be handy if, apart from 'Error = ', there would be a unique marker at the error locations to search for, rather than scroll until you found it. I often use '!!!' for that, but any marker with a low probablitiy of being present in the 'regular' output will do.

Ya - that is true. I can add that easily - in fact the position string is very likely to never have `!!!` in it so that  would probably be an easy place to add it. 

> 2 It would also be handy if, when optionally there would be a possibility to hide output lines and headers that contain no error. Giving all correct lines and headers with error count 0 a special css class and then having 2 buttons on top of the page to switch those from visible to hidden or something like that would probably do the trick.

Ya - I agree. I wanted to do this but ran out of time yesterday. Should not require much effort.

PR #190, opened 2016-11-18
Autotester Browser View Update #2

RCL-Carl:
Hi, 

As we discussed in the last PR for the autotester browser view, the latest HTML generation has a couple of short comings: 

1. Lack of ability to quickly search for tests with errors ("!!!" prefix missing)
2. Ability to expand/contract portions of the table

This PR fixes both of these short comings. I currently have it setup to collapse by default any module where all tests successfully completed. The user can then expand or collapse by clicking on the blue header for that module. 

Note that the code in this file is getting pretty spaghetti-like. It could use a code cleanup but for right now I don't want to spend any more time on it.

### DIFF
--- a/transcrypt/modules/org/transcrypt/autotester/__init__.py
+++ b/transcrypt/modules/org/transcrypt/autotester/__init__.py
@@ -130,6 +130,7 @@ class AutoTester:
 		self.excArea = "exc-area"
 		self.excHeaderClass = "exc-header"
 		self.collapsedClass = "collapsed"
+		self.modCollapseClass = "mod-collapsed"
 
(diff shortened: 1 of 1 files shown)


RCL-Carl:
Also you should be able to close Issue #184 with this commit.

JdeH:
**I've tried it out with some deliberate errors and I think it works fine! Thanks for this good piece of work.**

About the spaghetti: I haven't looked at the new code close enough yet to have a judgement about that. But it makes me think about the process in general. The result of this thinking will be in issue #193.

After having drawn up that issue, I'll complete this comment.

Issue #98, opened 2016-08-06
Regular expressions, Python API, JS engine


JdeH:
Many people use regular expressions in their programs, but for me this unfortunately happens to be a very persistent blind spot. Somehow they don't fit into my brain, I can understand them but not remember them, despite many years of using them in quite complicated ways.

In Transcrypt the 'normal' JavaScript regular expressions can of course be used without restriction, as can any part of JS. However it would be nice to have a re (old) or regex (new) module with an API as close as possible to the API CPython provides.

Writing Python-style pattern matching algorithms directly in JavaScript or Transcrypt would be FAR to slow!!!! So under this API, **the re or regex module should use JavaScript's regular expression facilities (which are written in C, just like Python's) as its workhorse as much as possible**. This is a mandatory requirement for any regular expression module suitable for real and heavy use. Also its footprint should be small, suitable for the niche that Transcrypt is aiming at: fast, small JavaScript code. So what's needed is a thin wrapper around JS, also in this case. Pareto's rule should be used as a guideline: Just leave out that 10% functionality that costs you 90 % of the effort and bloats the library in addition.

This is a labor-intensive, difficult job! Such a module would have to be VERY reliable and an extensive autotest (back to back with CPython, as any Transcrypt autotest, see docs) should come with it. It's ok, probably even advisable, to **start out with limited functionality**, assuming that it's clearly documented what's in and what isn't. But what's there should be rock solid!

I am aware that there are at least three Python-based RE modules out there, one for RapydScript (NT), one for RPython and one for Brython. If you use any code written by someone else, please respect licences, include copyright notices etc.! People should be credited for what they create. If you'd rather make a fresh start,  however tiny, that's fine as well.

To minimize logistic and license compatibility problems, the Transcrypt regular expression module should be distibuted using PyPi, rather than be part of the Transcrypt distribution, and have its own GitHub repository. However a link to it can be placed on the Transcrypt site to let developers know it exists.

Having efficient, fast CPython-style regular expressions available in a compact module would be of great value to Transcrypt developers.

AXGKl:
Can just comment, I'm in the same group than you unfortunatelly (user when I have to but not really understander...)
Wanted to add here a good site to compare python and js, feature wise:

http://www.regular-expressions.info/refcharclass.html

In fact they seem to be quite similar in features.

JdeH:
Indeed, that's why I hope that behind a CPython re api there can be an (efficient) JS re implementation.

[...]

PR #197, opened 2016-11-20
Initial Regular Expression Integration

RCL-Carl:
Hi, 

The primary thrust of this PR is to address issue #98 - the python re module. This PR also contains a few commits that update the autotester code base. The most important of these is a bug fix that corrects an error in the autotester where false positives were possible if a test suite contained only a single module and had an exception. Additionally, there is a single commit that fixes the KeyError exception (it was in \_\_standard\__.py but not in \_\_builtin__.js) and adds new exceptions from the standard builtin exceptions list to support the re module.

For the Python regular expressions module, I've taken the approach discussed in the Issue #98. The re module has two modes that can be controlled via a flag. The first mode translates python re syntax into javascript syntax and attempts to mimic python as much as possible leveraging @GULPF 's translation engine. This is the default mode. The second mode uses a strict javascript syntax mode to implement the python re API. This second mode has some limitations (named captures, etc don't exist in javascript), but it is functional for most simple regular expressions. 

I've added a base set of unit tests for both modes. These unit tests don't cover everything but they cover a fair amount of the API's basic structure. Note that not all tests pass at this point, due to bugs in the re module as well as bugs in transcrypt core. I have not included the re unit tests in the travis CI yet.

@GULPF - in this PR, I've basically just copied and pasted the code from your repo into the re module including license and referenced it directly as a first pass. This is obviously not optimal as there a lot of duplicate portions of the API in your implementation and in mine, but this at least gives us a starting point that we can build on. Additionally, the license for this code is MIT which I believe is more liberal than the Apache license so it is ok to distribute in this manner but we should likely discuss.

### DIFF
--- a/transcrypt/development/automated_tests/re/autotest.py
+++ b/transcrypt/development/automated_tests/re/autotest.py
@@ -1,0 +1,11 @@
+import org.transcrypt.autotester
+
+import basic_pyre
+import basic_jsre
+
+autoTester = org.transcrypt.autotester.AutoTester ()
+
+autoTester.run (basic_jsre, 'Basic JS RE Tests')
+autoTester.run (basic_pyre, 'Basic Python RE Tests')
+
+autoTester.done ()
--- a/transcrypt/development/automated_tests/re/basic_jsre.py
+++ b/transcrypt/development/automated_tests/re/basic_jsre.py
@@ -1,0 +1,30 @@
+from org.transcrypt.stubs.browser import __pragma__
+import re
+__pragma__("skip")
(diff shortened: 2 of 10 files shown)


GULPF:
Looks good! About the license issues - I've now switched to apache v2 in the rescrypt repo which should eliminate any trouble.

JdeH:
Since re can now carry the same license and this all looks quite good, lets indeed make this a standard module that's part of the distribution rather than a separate pip-installable package.

Highly pleased
Jacques

Issue #174, opened 2016-11-15
Python logging module


RCL-Carl:
Is anyone working on a python logging module ? I'd like to use the same interface that can target the web console but also for targeting other types of end points (like web sockets). I'd be interested in working on developing something like this. Would this be accepted as a module if I put something together ?

JdeH:
No one is currently working on a logging module, so it would be very useful if you picked that up.
From what I've seen from your `re` work, I expect that if you write a `logging` module, it will have a high chance of being accepted for the distribution.

### The criteria for acceptance of a module as part of the distribution are:

**1. Fits the design goals of Transcrypt**

Clarity, speed, compactness and robustness are favored over 100% completeness. This means a module reliably covering the most common usecases without being bloated is a good candidate for inclusion.

**2. Has an [autotest](http://sterlicht.alwaysdata.net/transcrypt.org/docs/html/autotesting_transcrypt.html)  coming with it**

The reason for this is that the core of Transcrypt still is in flux, and with every modification we must make sure nothing has fallen over, i.e. that the core modifications didn't break any module that is part of the distribution. The only way to make this feasible with limited resources is automated back to back testing with CPython.

I've seen you already know this, but for other contributers who happen to read this: You'll find examples of autotests in https://github.com/QQuick/Transcrypt/tree/master/transcrypt/development/automated_tests. You can just add a `logging` folder there, as @AXGKI did for the `time` module that he contributed.

The simplest way to make the logging module suitable for autotesting would probably be to add a handler to it that would pass logger output to  autoTester.check. That could e.g. be achieved by passing the autoTester instance to the handler and then have the handler call the `check` method for each message.

**3. Has some documentation coming with it**, as far as it deviates from the CPython version of that module.


### Location of modules

**1. Any 'standard' module that becomes part of the distribution** will be in https://github.com/QQuick/Transcrypt/tree/master/transcrypt/modules. This prevents name conflicts with CPython. So if your module is called 'logging', this is where it should go since there's also a CPython module by that name.

**2. In the rare case a module should be part of the distribution but there's a name conflict**, use Transcrypt's URL based module system, as is e.g. done with module: https://github.com/QQuick/Transcrypt/tree/master/transcrypt/modules/com/fabricjs
which can be imported with:

`from com.fabricjs import fabric`

This requires that the contributer has a URL at his disposition. In general the import format would be:

`from <myTopLevelDomain>.<myDomain>.<anyThing> import <myModule>`

[...]

PR #205, opened 2016-11-22
modules: Initial implementation of the `logging` python module

RCL-Carl:
This commit includes an initial implementation of the logging module and addresses Issue #174. It
includes a basic unit test for the logger only. The `logging.basicConfig`
and `logging.config.dictConfig` methods are not tested at this point.

I've stubbed in placeholders for some additional logging.Handler objects
that I think could be useful in the future but these are not tested
yet.

The provided unit test inserts a new TestHandler object that calls the AutoTester `check` method. Part of the tests also inserts a StreamHandler and allows manual inspection of the console.log but I haven't thought of a good way to test that other than manually.

### DIFF
--- a/transcrypt/development/automated_tests/logging/autotest.py
+++ b/transcrypt/development/automated_tests/logging/autotest.py
@@ -1,0 +1,12 @@
+# Main unit test platform for the logging module
+
+from org.transcrypt.stubs.browser import __pragma__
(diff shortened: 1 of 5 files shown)


JdeH:
:)

AXGKl:
absolutely awesome, hats off.

Issue #196, opened 2016-11-20
Bug: for loop over iter()


RCL-Carl:
Hi, 

I'm working on the python re.finditer function implementation and there seems to be an issue with the code generation for the for-loop over an iterable object generated by the `iter()` function. 

Python Code: 
```
a = 0
vals = [1,2,3]
ret = iter(vals)
for m in ret:
	a += m
```

Javascript Code:
```
		var a = 0;
		var vals = list ([1, 2, 3]);
		var ret = py_iter (vals);
		var __iterable0__ = ret;
		for (var __index0__ = 0; __index0__ < __iterable0__.length; __index0__++) {
			var m = __iterable0__ [__index0__];
			a += m;
		}
```

The issue is that `ret` returned by the `py_iter(vals)` returns an object of the form: 
```
__PyIterator__ {iterable: Array[4], index: 0}
index:0
iterable:Array[4]
next:__jsUsePyNext__()
Symbol(Symbol.iterator):()
__proto__:Object
```

which gets assigned to `__iterable0__`but what we really want is for the value `ret.iterable` to be assigned to `__iterable0__`.

JdeH:
Hi,

Call to iter () requires iterator support on the JS side, so -e 6 switch. Minified code will run on es5 anyhow, through Google Closure polyfills.
Will put this in the docs!

KR
Jacques

PS Rationale: When the 'universal iterator model' was added to Transcrypt, es 6 was just released. Since Closure can generate es5 from es6 and the model was quite complicated in itself, I decided not to retrofit for es5.

RCL-Carl:
From what you have described, it sounds like this is something that can be determined at compile time, no? If so, I think it would be better if the compiler were to throw an error in this case. Currently, it silently completes the compilation phase and generates a js file with faulty implementation. I think updating the documentation is good too, but failing fast in the compilation stage is better.

PR #206, opened 2016-11-23
Functional re module - for the most part.

RCL-Carl:
This PR contains an update to the regex module that integrates the python translation and javascript native regex syntax modules. Currently, this code uses the 'tconv' pragma at the file level in `re/translate.py` which likely could be optimized once we have all the tests passing.  I've also added some more unit tests to cover additional part of the api.

@GULPF - the Python regex translation has a problem with patterns that have more than one named group. It seems that the namedGroups dict gets the first named group right but each one after that is one more than it should be? Not totally sure. The only other major outstanding issues is the finditer function which has an issue I've outlined in Issue #196. Currently, the js portion of the unit tests does not evaluate the python only regex syntax tests. The `Match.expand` is not implemented at this point. 

I've made an update to the autotester check methods that _should not_ affect its performance or call signature in the unit tests but we should test thoroughly. The CI works so I'm not too worried. This change allows the user to specify the ancestor level for determining which frame in the stacktrace to use for reporting in the browser. I've added a utility method called `checkEval` which when passed a function will evaluate the function and turn any thrown exceptions into errors. This required enabling kwargs on the check method so that I could pass an argument for the ancestor level.


JdeH:
@GULPF @RCL-Carl:

Thanks for a great  contribution!

I like the 'ancestor' kwarg!
Already noticed the underlying problem, but didn't even think about a solution yet.

GULPF:
@RCL-Carl If you don't wan't to mess with translate.py, I can fix it in a few days. There are some other issues in it as well (in particular, regex conditionals will mess up the capture group id's - needs to be handled in PyRegex).

PR #215, opened 2016-11-26
update to latest and Python warnings module initial implementation

RCL-Carl:
This commit updates the subject_libraries branch to the latest master and contains an initial implementation of the `warnings` module. This module has some limitations. Specifically, the `catch_warnings` method is not implemented. The user is encouraged to use the `warnings.warn_explicit` method  and not `warnings.warn` because right now the meta info provided by `warn` is not particularly useful.


RCL-Carl:
This PR targets Issue # 191

Issue #181, opened 2016-11-17
Bug with arrays and None in autotester


GULPF:
Seems like there is a bug with how the auto tester handles `null`. `autoTester.check(None)` works correctly, but `autoTester.check([None])` always fail.

JdeH:
While it is very useful and has caught countless bugs, the autotester isn't perfect yet. The main limitation is that small format deviations and accuracy issues can give errors. I've had to work around several of these limitations. I'll take a look into this one. At first sight I'd say it should be easily solvable, but I'm not sure.

RCL-Carl:
This is a bug in the `Array.prototype__repr__` method in `org/transcrypt/__javascript__/__buitin__.mod.js`

Here is an excerpt of `__repr__`: 

```
        for (var index = 0; index < this.length; index++) {
            if (index) {
                result += ', ';
            }
            try {
                result += this [index] .__repr__ ();
            }
            catch (exception) {
                result += this [index] .toString ();
            }
        }
```

This is repr over a list object (Javascript Array). It goes through each element in the array and tries to repr it, if that fails, it tries to call the `toString` method to convert to string. 

Both of these fail on `null` and the exception propagates up to the `repr()` builtin function's exception handler - where it attempts to check for `null` (\_\_builtin__.mod.js:284) but because it is a list of nulls, this check fails. 

I think there are two alternatives here. First, we could add a check before the `toString` method above to check if the element is null before calling to string. Something like: 

```
            try {
                result += this [index] .__repr__ ();
            }
            catch (exception) {
                if ( this [index] != null ) { 
                    result += this [index] .toString ();
                } else { 
                    return("None")
                }
            }
```

Alternatively, we could recursively call `repr()` here instead of trying to recreate the functionality of repr in another location

The first approach would certainly be simpler and less code.

PR #220, opened 2016-11-27
Fix Repr Issue #181

RCL-Carl:
This commit fixes an issue with the `repr` builtin function that was causing the represented strings not to show value values when None or null values were present. This issue triggered the finding of a couple of other issues that have subsequently been fixed by using a recursive call to `repr()`


RCL-Carl:
This addresses issue #181 (evidently this doesn't get linked properly in the title)

JdeH:
Well of course this should be recursive, since the datastructure represented itself can be recursive.
Seems so simple after the fact...

Issue #221, opened 2016-11-27
Dict Creation Issue


RCL-Carl:
In python one can create a copy of a dict by calling the following (this is obviously a contrived example but the concept of making a copy of a dict is relatively prevalent one): 

```
c = dict({ "a": 1, "b" : 2 })
d = dict(c)
d["a"] = 2

# Now 
# d["a"] -> 2
# c["a"] -> 1
```

In transcrypt, if one attempts to do this, the js runtime throws a `TypeError: Cannot redefine property: __class__` exception because the `Object.defineProperty` method is invoked on `c` to create dict `d`. Per the [documentation here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty), the properties added through this method are immutable.

I've noticed that sometimes it throws on redefinition of `__iter__` and sometimes on `__class__` - not sure why it isn't consistent.

JdeH:
Fine!

PR #225, opened 2016-11-27
Fix for Issue #221 - Dict Creation and other Bugs

RCL-Carl:
Hi, 

This PR fixes some bugs in the dict object creation and methods. This is primarily focused on fixing the bug in Issue #221 but after creating some additional unit tests, I found that there are some other issues that needed fixing. 

Please note, that this PR addresses dict creation on objects that are not using `Map` because these are not tested in Travis CI or `test_server.py` yet. We will need to adapt these changes to cover both cases. If you would like to wait on this PR until these fixes are incorporated to both variants please let me know.  

Additionally, this PR contains a commit addressing inconsistencies in the js for `__builtin__.mod.js` with respect to semicolons.

### DIFF
--- a/transcrypt/development/automated_tests/transcrypt/dictionaries/__init__.py
+++ b/transcrypt/development/automated_tests/transcrypt/dictionaries/__init__.py
@@ -44,4 +44,16 @@ def run (autoTester):
     autoTester.check (tel.pop ('guido', 1))
     autoTester.check (tel.pop ('edsger', 2))
     autoTester.check (tel.pop ('foo', 'bar'))
-    
+    autoTester.check (tel.pop ('foo', None))
+
+    # Check exceptions
+    knights = {'robin': 'the brave', 'gallahad': 'the pure'}
+    autoTester.check (
+        autoTester.expectException ( lambda: knights.pop("batman") )
+    )
+    autoTester.check (
+        autoTester.expectException ( lambda: knights.pop("batman", None) )
+    )
+    autoTester.check (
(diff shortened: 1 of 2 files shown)


JdeH:
Hi

I agree with commit f768bb3 and 62c31cc, these are good improvements of the dict and indeed address e.g. #221.

I only agree partially with commit 2ee439e. Adding a trailing semicolon is correct in this case, since the whole thing is indeed an assignment statement:

```
var chr = function (charCode) {
    return String.fromCharCode (charCode);
 }; 
```

But adding a semicolon in this case is out of place, since the } here merely terminates a block of statements but isn't part of a statement in its own right:

```
function __JsIterator__ (iterable) {
      this.iterable = iterable;
      this.index = 0;	
}
```

Adding a ; here would just add an empty statement, which is pointless.
Since this is a minor thing, I'll merge anyhow and correct it later.
Let me know if you think I misinterpret JS syntax rules.

KR
Jacques

RCL-Carl:
Hi,

Sure - your reasoning on the semicolon post function definition makes
sense. In the file as it existed, both cases were present, ie, some
functions had semicolons and others didn't, so I just picked one and went
for consistency. I'll remove them the next time I touch that file if you
haven't done so already. Less characters in a file the better I say.


On Mon, Nov 28, 2016 at 12:54 AM, Jacques de Hooge <notifications@github.com

JdeH:
Maybe it's already fixed, haven't looked too closely.

Issue #193, opened 2016-11-19
T-RFC1: Cooperation and division of labour.


JdeH:
[EDIT]

------

As a result of this RFC, the discussion below and experiment that followed it, @JdeH decides the following:

> 1. To keep the subjects and the subject owners, since without them the scope of my mind is too much the limiting factor. My impression is the subject/owner thing really speeds up the project and enhances its quality. While maintaining course, I will base myself largely on the judgement of the subject owners for details.
> 2. Keep the janitor team with push privilege to the whole repo. currently it's having me as only member, but that may be altered simply by adding other team members, on temporary or lasting basis if the need arises. Let's cross that bridge if we get to it, currently I don't think it's a problem.
> 3. To create labels to assign issues to subjects. Such a label is a request from the janitor to the subject owner to take the issue under his (m/f) care.
> 4. To keep the subject teams, simply as an explicitation of who the subject owners are.
> 5. To do away with subject branches. Nice try but didn't work as expected.
> 6. To NOT introduce subtree's, submodules, subrepo's or whatever they may be called, since they tie directory structure to subject structure, which is far too restrictive. File level permissions maybe come closest to what we really need, but GitHub currently doesn't allow that.

------

[END EDIT]

Transcrypt: the transition from chapel to bazaar
------------------------------------------------

Now that more people are getting involved in developing Transcrypt, I'm looking for a way to allow this project to grow orderly without myself being a bottleneck. This means that I'll have to let go without losing control.

What I would like to do is to take some parts of Transcrypt, I'll name them subjects, and assign them to  an owner. Any changes to a particular subject go through the owner for approval before being commited. The owner takes responsibility for the code quality of that subject, and it is upto him/her to decide if an when to refactor. I'll stay in the loop as far as code functionality is concerned, and I'll also do the final merges for now, but I will not impose my ideas on code structure upon every part of Transcrypt.

Ownership: coordinated freedom
------------------------------

When I work on a piece of code alone, it's structure is like a sawtooth. When adding a new feature, I primarily try to get it to work, structuring my code as well as I can, but sometimes postponing reorganization. Until I discover that the sub-optimal structure starts to interfere with my productivity. I then stop adding features and concentrate on refactoring for a while.

[...]

PR #226, opened 2016-11-27
Master

AXGKl:
Hi, test_flags parametrisable now. Default on travis still without -e 6 since I have test errors (!)


JdeH:
Hi @AXGKl 

I've tried to summarize the status w.r.t. the subjects and owners in an [EDIT] at the top comment of #193.

What it boils down to is the following:

- Subjects and owners are useful, even necessary to allow Transcrypt to grow unimpeded.
- Submodules, subtrees and subrepos are all equally unuseable since they tie the directory structure to the division in subjects, which is too restrictive. Github shouldn't dictate our dir structure.
- The experiment with subject branches and push priv to them for the subject owner made the version trellis much unworkably complex. I must say I completely lost track within a day or two :(
- File level perms is what we really need, but Github doesn't support them.
- There's janitor team that can push to the master branch.
- Any team member can make subject branches, but can also work from personal branches.

Currently I am the only member of the janitor team. In general making all subject owners member of the janitor team carries a risk, since that would mean that a subject owner of e.g. documentation accidentally could push to the core. However should the need arise, since it's a janitor _team_ rather than a janitor _person_, the janitor function can be shared or transferred on a temporary or permanent basis.

If file level perms are added to Github (will they ever be?) I would prefer allowing subject owners to push to the files of the subjects they own.

Currently I am looking into Transcrypt almost on a dayly basis, and I can even easily merge things from my mobile phone, so I think for now this will work. What do you think?

JdeH:
BTW I'll look into the test failures.

AXGKl:
You know my only fear is that you burn out and then I have not Transcrypt anymore. As long as you say *you* have no probs to do all the logistics yourself then *I* have no prob as well.

AXGKl:
Btw: File by File tracking like in svn we'll never have in a DVCS, they all work tracking the state of directories and changesets, but not file by file, by design. They could not offer the performance otherwise.
---

I personally would not have big trouble with the subject <-> directory tree mapping, one could argue its a feature forcing us to keep seperate things seperate, but thats a personal opinion ;-)

JdeH:
Currently PR's are of high quality and well documented, and I rely on the judgement of the subject owners to a large extent, so accepting PR's doesn't cost me much time or effort. But still you touch upon an important subject on which it's necessary to have clarity.

You currently are one of the early adopters using TS for serious work. I am very aware of that fact, and the demands that this poses to continuity. I want to reiterate that, while I am the initiator of Transcrypt, I have no special rights with regard to it. It is open source, and while I try to serve the project, it doesn't depend on me for its existence.

After the subject branch experiment I have now returned to concentrate upon the core, since that's the part I am best at. I can imagine that at some point the janitor function will take too much of my time. I see my role as a coordinating one, and if somehow I can not fulfill that role properly for a shorter or longer period, it's just a matter of augmenting or swapping the janitor team.

PR #227, opened 2016-11-28
__name__ and __filename__ macro features

RCL-Carl:
Hi, 

This PR builds on the work for `__file__` and `__line__`. This adds a compile-time replacement macro for `__name__` to insert the module name like in python. 

Additionally, this adds a `__filename__` macro that includes the name of the file without the absolute path. The `__filename__` macro will include the subdirectory if the filename starts with `__init__` because I don't think that `__init__.py` would be useful to anyone.


JdeH:
- Adding the `__name__` and `__filename__` macro: Good idea, let's have that. I thought about several alternative for the `__filename__` macro but I think it's the most simple and straight solution.

- Removing trailing whitespace: OK, but I hope my editor (Notepad++) doesn't ruin it again, will see if I can find a setting avoid trailing ws.

PR #230, opened 2016-11-28
test flags should now work

AXGKl:
test flags did not work.

I now committed the -e 6 flags as well RESULTING IN BROKEN TRAVIS


AXGKl:
hi, test flags fixed. I verified that the -e6 tests are really broken currently in div fixes -> travis is broken currently.

JdeH:
I think I've broken -e 6, almost fixed.
Will accept this PR first.

PR #233, opened 2016-11-28
-e 6 now works on travis

AXGKl:
Hi, so with new firefox 49 all works.

thanks for the hint with -r -e 6 !!


JdeH:
Nice piece of work.
This will save me quite some time in the dev. cycle.

PR #237, opened 2016-11-29
Updates to re module

RCL-Carl:
This PR contains some bug fixes to the re module. Additionally, I've modified the `re.finditer` method handling to only work with es6 and later as es5 does not support iterators (per Issue #196). The unit tests will now run in both es6 and es5. Please note that we have one outstanding issue which seems to be a short coming in the conditional group handling.


JdeH:
Well, again some steps closer!
I'll leave #98 open because of the outstanding issue.

PR #238, opened 2016-11-29
Logging Config Baseline Implementation

RCL-Carl:
This PR contains updates to the the `logging` module to add basic functionality for `basicConfig` and `dictConfig`. I've added basic unit tests for these features and I've additionally added some unit tests for the buffering handlers.

This PR also contains a minor update to the `re` module to update it to the latest support of `callable` which wasn't available earlier.

This commit primarily addresses issue #174

### DIFF
--- a/transcrypt/development/automated_tests/logging/autotest.py
+++ b/transcrypt/development/automated_tests/logging/autotest.py
@@ -4,9 +4,15 @@ from org.transcrypt.stubs.browser import __pragma__
 import org.transcrypt.autotester
 
 import logger_tests
+import config_tests
+import basicConfig_tests
+import buffering_tests
 
 autoTester = org.transcrypt.autotester.AutoTester ()
 
 autoTester.run( logger_tests, "logger_tests" )
+autoTester.run( config_tests, "config_tests" )
+autoTester.run( basicConfig_tests, "basicConfig_tests" )
+autoTester.run( buffering_tests, "buffering_tests")
 
 autoTester.done()
--- a/transcrypt/development/automated_tests/logging/basicConfig_tests.py
+++ b/transcrypt/development/automated_tests/logging/basicConfig_tests.py
@@ -1,0 +1,36 @@
+# This file contains some unit tests for the logging.basicConfig
+# method.
+
+from org.transcrypt.stubs.browser import __pragma__, __envir__
+import logging
+from utils import TestHandler, resetLogging
+
+def run(test):
+    """
+    """
+    resetLogging()
+
+    basehdlr = TestHandler(test, 5)
+    fmt = logging.Formatter(style="{")
+    basehdlr.setFormatter(fmt)
+
+    logging.basicConfig(handlers = [basehdlr], level=logging.INFO)
+
+    root = logging.getLogger()
+    test.check(root.hasHandlers())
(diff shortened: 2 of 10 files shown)


JdeH:
What stage would you say the logging module is in?

RCL-Carl:
I think most of the core functionality is there. I think it would help if it got a little broader testing, both by other users and unit tests. The dictConfig and basicConfig unit tests are pretty shallow at this point. There is basically no coverage of the "cfg:\\..." syntax - I'm not really sure I see a significant use case, maybe it should be removed. 

I have implemented an AJAX-based logging handler but I don't have a good way to unit test it with autotester. I've created a basic manual unit test for GET/POST and that seems to work. I haven't started on a websocket version yet.

JdeH:
About the cfg://... syntax: If you have serious doubt as to wether it's working, probably outcommenting that code, with a remark that it's ok in principle but not yet tested is best. If it turns out that people use it after all, it can be reactivated and tested. That's what I think. But you do what you think is best.

About testing the AJAX based logging handler: great that it's there! And about testing it: There are more manual tests, so just add it to that directory for now. In the long run maybe some effort must be made to "promote" tests from manual to automatic. E.g. in the shipment test all demo's are tested just by clicking and eyeballing. It's doable for now, but there are tools that can automate this. But let's not now invest too much effort in that. Ten minutes (or even half an hour) of clicking before a release isn't a big problem. A manual test is far better than no test at all.

What is important for manual tests (and I haven't yet done anything about it) is to describe what to test, how to test and what the results should be, so that we can go through them systematically.

AXGKl:
@JdeH: Added the logging module to travis (in set1): https://github.com/QQuick/Transcrypt/pull/239


@RCL-Carl:

> but I don't have a good way to unit test it with autotester

When I'm testing a module, I'm in the directory of it, and could check for the presence of meta data about what to import into the test server process, potentially adding websocket handlers or ajax endpoints.
For example by importing classes from modules and calling register functions passing the test server app, so they can extend the routes known to the test server.

Those classes could be based on the client sided classes btw, I'm doing this like:

```python
class Foo:
     ''' client sided attrs and methods '''
     def foo(self): ...

__pragma__('ifdef', 'on_server')
class Foo(Foo):
     ''' server only stuff ''''
    @classmethod
     def register(cls, app):
             ....

__pragma__('endif')

```
The registration could call a helper which knows how to register endpoints for flask, bottle, what not?

thoughts?

RCL-Carl:
> About the cfg://... syntax: If you have serious doubt as to wether it's working, probably outcommenting that code, with a remark that it's ok in principle but not yet tested is best. If it turns out that people use it after all, it can be reactivated and tested. That's what I think. But you do what you think is best.

Ya - I'll have it throw NotImplementedError and if someone complains we can deal with it then. 

> When I'm testing a module, I'm in the directory of it, and could check for the presence of meta data about what to import into the test server process, potentially adding websocket handlers or ajax endpoints.

Yes, I think that there is a solution that could make this work, but the complexity that would have to be added to `test_server.py` and `autotester` to make this work are more than it is worth at this point. I'm with @JdeH that a manual test is sufficient for now and if someone is ambitious they can integrate it later.

PR #240, opened 2016-11-30
Manual test for AJAX handler and cleanup of config

RCL-Carl:
This PR contains an update to the AJAX logging handler and a manual test for checking this feature. I've also made the converter syntax throw `NotImplementedError` as discussed because this code in `logging.config` is not well tested. 

I think this commit completes the basic necessities for the logging module and closes issue #174 . There are probably bugs and things we will have to fix but I think this gets us to a point where we can at least start testing it.

### DIFF
--- a/transcrypt/development/manual_tests/module_logging/README.md
+++ b/transcrypt/development/manual_tests/module_logging/README.md
@@ -1,0 +1,50 @@
+Logging AJAX Handler Manuall Test
+=================================
+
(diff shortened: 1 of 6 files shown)


JdeH:
I'll try the manual test as soon as I find the time.
Currently starting up a new project for a customer, so a bit thinly spread.
Will get back to this, probably in the weekend.

PR #248, opened 2017-01-03
Add js_global to aliases

daboross:
'global' is a valid symbol in JavaScript, but not python. Add 'js_global' as an alias to 'global' to allow for accessing this.


JdeH:
Right!

PR #271, opened 2017-01-31
Fix `from x import y` inside functions.

### LINKED ISSUE #269: from x import y statements cause errors when used within a function
I believe this is due to Transcrypt trying to export modules imported only within a specific function?

Test code: `inner/__init__.py` is empty.

`inner/moda.py`:

```python
from inner import modb

def modafunc():
    return "stuff"


def both():
    return modb.modbfunc()
```

`inner/modb.py`:

```python
def modbfunc():
    from inner import moda
    return moda.modafunc()
```

`main.py`:

```
from inner import moda

print(moda.both())
```

Output:
```
/home/daboross/Projects/Python/TranscryptTest/target/main.js:2080
						__all__.moda = moda;
						               ^

ReferenceError: moda is not defined
    at Object.__init__ (/home/daboross/Projects/Python/TranscryptTest/target/main.js:2080:22)
    at __init__ (/home/daboross/Projects/Python/TranscryptTest/target/main.js:42:28)
    at Object.__init__ (/home/daboross/Projects/Python/TranscryptTest/target/main.js:2047:17)
    at __init__ (/home/daboross/Projects/Python/TranscryptTest/target/main.js:42:28)
    at /home/daboross/Projects/Python/TranscryptTest/target/main.js:2088:14
    at main (/home/daboross/Projects/Python/TranscryptTest/target/main.js:2096:5)
    at Object.<anonymous> (/home/daboross/Projects/Python/TranscryptTest/target/main.js:2099:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
```

Generated `main.js`: https://gist.github.com/daboross/2ae5db3f25356372c00253f551c93754

This does not occur if `moda.py` and `modb.py` are top-level modules, and only `import modb` is used. However, with that setup, `moda` is still exported, just as an empty object!

daboross:
This should fix #269. As a side affect, this also allows cyclical imports to work correctly when one of the imports is done inside of a function.

### DIFF
--- a/transcrypt/development/automated_tests/transcrypt/modules/__init__.py
+++ b/transcrypt/development/automated_tests/transcrypt/modules/__init__.py
@@ -54,3 +54,7 @@ def run (autoTester):
     autoTester.check (mod3Hundred)
     autoTester.check (mod3GetTwoHundred ())
     autoTester.check (A (123.321) .f ())
+
+    from modules import mod4
+    # From ... import within function, with circular import
+    autoTester.check(mod4.mod4Add2FromMod5(449))
--- a/transcrypt/development/automated_tests/transcrypt/modules/mod4.py
+++ b/transcrypt/development/automated_tests/transcrypt/modules/mod4.py
@@ -1,0 +1,6 @@
+def mod4Add2FromMod5 (variable):
+    from modules.mod5 import mod5Add2
+    return mod5Add2 (variable)
+
+def mod4Add1(variable):
+    return variable + 1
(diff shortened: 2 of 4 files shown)


daboross:
Still working on a test for this, not entirely sure how to do that but I'll try to add one.

daboross:
If there's anything you'd want differently in the tests or anything in addition to be added, I can do that. I'm still not entirely sure I understand the auto testing, but I tried to add this in somewhere that seemed right.

JdeH:
Elegant, making the gap with CPython even semantics smaller.

Issue #326, opened 2017-05-16
How to use `zip()` on custom iterable?


theodox:
It looks like transcrypt expects a  `map()` method on custom iterables to make them work with zip().  If I just implement something like this:

``` 
class Vector:
     def __init__(self, *val):
         self.values = val

     def __add__(self, other):
        return Vector(*(a + b for a, b in zip(self, other)))
   
    def __iter__ (self):
       for item in self.values:
            yield item
```
and try something like 

```
Vector(1,2) + Vector(3,4)
```
or 

``` result = zip(Vector(1.2), Vector(3,4))
```

It will fail with `shortest.map is not a function`.    What would have to be implemented for this to work?

withrocks:
I'm experiencing the same problem without explicitly using zip.

I'm using Python's `enumerate`, which is under the hood implemented as a call to `zip`.

withrocks:
The issue I was having seems to have had to do with zipping a list of integers (or something else) with a string as the second argument, e.g. when calling `enumerate('abc')`.

I believe I've fixed it for my particular case, I'm not sure that it fixes the issue @theodox had, since I ran into another error when trying to reproduce that.

PR: https://github.com/QQuick/Transcrypt/pull/327.

withrocks:
@theodox: The pull request has been merged. It would be interesting to know if the patch solved your issue too.

JdeH:
@theodox 

The zip function has not yet been made suitable for the use of iterators.

As a workaround the following solution could be used here:

```
class Vector:
    def __init__(self, *val):
        self.values = val

    def __add__(self, other):
        return Vector(*(a + b for a, b in zip(self.values, other.values)))
            
__pragma__ ('opov')
v = Vector(1,2) + Vector(3,4)
__pragma__ ('noopov')

print (*v.values)   

```

BTW as you already may know,  iterators in general require the `-e 6 `switch

JdeH:
This will stay a workaround for the foreseeable future...

PR #327, opened 2017-05-20
Fix zip function when zipping characters of a string with another list

withrocks:
This pull request should fix an issue with zipping characters of a string with other lists, see #326 (second comment.)

The zip function was only checking if the first argument was a string and splitting it in that case. This meant that if the first argument was e.g. a list of integers and the second was the string 'abc', (as in the case of `enumerate('abc')`), the zip function didn't work. The patch splits all arguments to zip that are strings.

Includes a minor fix for .gitignore, where my `egg-info` was not being ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 **.__pycache__/
 dist/
 transcrypt/docs/sphinx/_build/
-transcrypt.egg-info/
+*.egg-info/
 transcrypt/development/attic/
 transcrypt/development/docs/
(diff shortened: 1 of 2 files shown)


withrocks:
I did not add a test for this as I wasn't sure where/how you would want it, but one is included in this branch: https://github.com/withrocks/Transcrypt/tree/326-zip-iterable-with-tests/transcrypt/development/manual_tests/test_326

PR #335, opened 2017-06-08
Correct a typo and add `.. note::` to `installation_use.rst`

all3fox:
Also rephrase one particularly awkward sentence in `philosophy_directions.rst`

Thank you for `transcrypt`! Also, my editor `emacs` was opening those `.rst` files in `DOS` encoding, you probably should convert them to `UTF-8` or something?


JdeH:
Thanks!

PR #340, opened 2017-06-22
Add empty transcrypt module __future__

all3fox:
The `__future__` module allows to use things like `print()` function in python `2.x` code. Currently, python code that goes above and beyond to work on both `2.x` and `3.x` uses this module heavily. 

AFAIK transcrypt could still transpile such code treating it like `3.x` code. For instance, the simple:
```

from __future__ import print_function

print('Hello, world!')
```

could be made to work just by shadowing the `__future__` module with an empty `transcrypt`'s `__future__` module.

What say you? I am also looking at how `transcrypt/development/*tests` are organized but I have not yet wrapped my head around as to how to write some tests here.


JdeH:
Good idea!

PR #341, opened 2017-06-24
Improve documentation about autotesting

all3fox:
In my quest to write some tests for my [`__future__` PR][1] I decided to slightly improve the documentation about the basic idea behind `autotest` and its `hello` example.

[1]: https://github.com/QQuick/Transcrypt/pull/340


JdeH:
Useful additions, thanks!
The autotest process could indeed do with better docs and the hyperlink you made.

Some minor nits. (transcrypt -> Transcrypt, javascript -> JavaScript).
Admittend JavaScript isn't standard spelling, but it is what's done in the rest of the docs.
Of course officially it's called ECMA script, but that sounds a bit to formalistic to my taste.

KR
Jacques

PR #344, opened 2017-06-24
Update bottle to 0.13-dev, use python3 in run and test_server

all3fox:
Using the `dev` version of bottle just because I copied it straight from github. It works, so no worries.

Also, I just "fixed" the `test_server.py` script without getting into detail as to why it needs to `reload` the `sys` module (it most probably does not). Again, it works in python `3.x` just fine.

### DIFF
--- a/transcrypt/development/continuous_integration/bottle.py
+++ b/transcrypt/development/continuous_integration/bottle.py
@@ -2,71 +2,116 @@
 # -*- coding: utf-8 -*-
 """
 Bottle is a fast and simple micro-framework for small web applications. It
-offers request dispatching (Routes) with url parameter support, templates,
+offers request dispatching (Routes) with URL parameter support, templates,
 a built-in HTTP Server and adapters for many third party WSGI/HTTP-server and
 template engines - all in a single file and with no dependencies other than the
 Python Standard Library.
 
 Homepage and documentation: http://bottlepy.org/
 
-Copyright (c) 2013, Marcel Hellkamp.
+Copyright (c) 2017, Marcel Hellkamp.
 License: MIT (see LICENSE for details)
 """
 
-from __future__ import with_statement
+import sys
 
 __author__ = 'Marcel Hellkamp'
-__version__ = '0.12.9'
+__version__ = '0.13-dev'
 __license__ = 'MIT'
 
-# The gevent server adapter needs to patch some modules before they are imported
-# This is why we parse the commandline parameters here but handle them later
+###############################################################################
+# Command-line interface ######################################################
+###############################################################################
+# INFO: Some server adapters need to monkey-patch std-lib modules before they
+# are imported. This is why some of the command-line handling is done here, but
+# the actual call to _main() is at the end of the file.
+
+
+def _cli_parse(args):  # pragma: no coverage
+    from argparse import ArgumentParser
+
+    parser = ArgumentParser(prog=args[0], usage="%(prog)s [options] package.module:app")
+    opt = parser.add_argument
+    opt("--version", action="store_true", help="show version number.")
+    opt("-b", "--bind", metavar="ADDRESS", help="bind socket to ADDRESS.")
+    opt("-s", "--server", default='wsgiref', help="use SERVER as backend.")
+    opt("-p", "--plugin", action="append", help="install additional plugin/s.")
+    opt("-c", "--conf", action="append", metavar="FILE",
+        help="load config values from FILE.")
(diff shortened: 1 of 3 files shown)


all3fox:
By the way, how about putting it into `requirements.txt` and pip installing it on Travis instead of keeping the `bottle.py` file in your repo? Can do that no problem.

JdeH:
@all3fox
This touches *SUB: continuous integration* so I'd like the opinion of the subject owner on this.
@AXGKI, I think positively of this pull request, how about you?

(And by the way, @AXGKI, how are you these days?)

all3fox:
Personal opinion: as python developer, one would expect either `pip install -r requirements` and then `python setup.py test` or `pytest` or `nosetests` or `tox` or whatever "classic" to do the job. Instead, you really have to read `.travis.yml`, then `README.md` in `development/continuous_integration` and then (which is the best actually) find the [transcrypt youtube video][1] which shows you how things work. That is too slow/discouraging for newcomers.

Some tests probably could do away with `node.js` instead of a full-blown browser environment. If you know how to set such things up (you run python code, then run js code, then compare the results. Preferably inside a known framework like `pytest` or `unittests`), please explain the basic components/layout.

The idea here is, for instance, how [jquery does it][2]: they have a healthy amount of their test base that just runs with a conventional `npm run test` and then, once you really need it, you can setup the LAMP (or [whatever stack](https://github.com/jquery/jquery#running-the-unit-tests)) to run the more involved unit tests.

[1]: https://www.youtube.com/watch?v=lvTjKWVOwXA
[2]: https://github.com/jquery/jquery/blob/36389288e8c7f2538b5ae7dd6877e81a9a76c644/package.json#L63

AXGKl:
Hey guys, I'm totally sorry for my silence, got pretty involved this year in other stuffs. In general I'm well though, thanks for asking Jacques @JdeH , glad to see Transcrypt does so well, it deserves it!!

@all3fox: Your personal opinion is for sure justified, the whole thing was a one night hack to have a real browser running through all the autotests of Jacques. Meaning: All was basically there, just the browser automation was missing, thats why I wrapped some javascript around them, to have the browser hit one test after the other and served them from bottle. Its awfully non standard but it got us over the important start of the second phase of the project when more people got involved with PRs.

I honestly don't know how far you get w/o a real browser, I just know I tried with phantomjs and failed miserably - but I'm really not good in javascript, maybe there are better possibilities these days.

All in all: I think in this phase of Transcrypt a rewrite of the test framework would be justified. Alone - this year I won't be able to contribute, timewise I'm afraid, but for sure I can explain you what I did in detail, if you intend to rewrite my stuff.

the1337guy:
Well you don't need JS to run PhantomJS at all anymore.. Maybe use Selenium? That's why it was made for ? :P

AXGKl:
@the1337guy : Yes, selenium would be the tool to go for - once we have browser js compat errors (do we?) - but does not get you around the need for the test server, guiding the browser through the autotests, no?. And getting firefox up in headless mode in travis was not that big an issue, so I took that road.

the1337guy:
Well, Selenium does have error wrapping services plus you might not use Firefox at all, PhantomJS can be used with Selenium.

all3fox:
Ok, let's put a pin into the Selenium idea and into the "some common test framework" idea.

I would like to suggest to focus on the immediate issues this PR addresses:

1. It updates `bottle` framework and makes it use `python3`
2. By doing that, it eliminates the need to juggle between a `python2`-driven virtual environment to run tests and a `python3`-driven virtual environment to develop Transcrypt.

Merging this PR will make at least my experience more pleasant and should not break the existing workflow for others.

AXGKl:
@all3fox: Alex I see me=`was requested for review` - what exactly you want here? I mean the print functions should work. Regarding bottle I fear I can't help much, in my company (and industry) we are still at py2, have no experience with 3. What I can is try explain you the mechanics, you're requesting that?

all3fox:
Hey @AXGKl to be honest I would like you to just say that this PR makes sense. Also, if you take a closer look, you will see that it was @JdeH who requested the review:

https://github.com/QQuick/Transcrypt/pull/344#event-1137701145

Finally, your youtube video explains a lot and makes a lot of sense, so I do not have any questions :)

Finally-finally, to accept a review (since it was requested of you) you need to follow these guidelines: https://help.github.com/articles/approving-a-pull-request-with-required-reviews/

Basically, you can just do steps 5 and 7 from that guide even without leaving any messages.

PR #345, opened 2017-06-24
Add more (automatically generated) places to .gitignore

all3fox:
When you experiment with transcrypt, its repo quickly gets more and more autogenerated things. This is an attempt to tighten up the `.gitignore` file slightly so as to reduce the amount of noise in `git status`

Refs: https://github.com/QQuick/Transcrypt/issues/342


JdeH:
Quite useful.

PR #347, opened 2017-06-25
Add tests for transcrypt module __future__

all3fox:
Here are some promised tests for the `__future__` module that you merged earlier. (This PR message is a superset of the commit message). I am using the `0.13-dev` version of bottle (and still haven't figured out/don't want to figure out how to run two pythons in one virtual environment), so we will have to see if this tests pass CI on current `2.x` and `0.12.9` Travis (I've checked them locally on `0.13-dev` with `3.6.1` python)

The two things that I could not make to work in tests: byte strings with `b'Hello, world!` syntax and the `with` statement with the custom simple context manager. I left them commented out. Any thoughts?

Add simple tests for these features:
- division
- generators
- nested_scopes
- print_function
- unicode_literals

Tests that need attention:
- byte strings in unicode_literals
- with_statement (simple Context Manager does not work)

Tests that were not considered:
- absolute imports

### DIFF
--- a/transcrypt/development/automated_tests/__future__/autotest.py
+++ b/transcrypt/development/automated_tests/__future__/autotest.py
@@ -1,0 +1,19 @@
+import division
+import generators
+import nested_scopes
+import print_function
+import unicode_literals
+# import with_statement
+
+from org.transcrypt.autotester import AutoTester
+
+atester = AutoTester()
+
+atester.run(division, 'division')
+atester.run(generators, 'generators')
+atester.run(nested_scopes, 'nested_scope')
+atester.run(print_function, 'print_function')
+atester.run(unicode_literals, 'unicode_literals')
+# atester.run(with_statement, 'with_statement')
+
+atester.done()
--- a/transcrypt/development/automated_tests/__future__/division.py
+++ b/transcrypt/development/automated_tests/__future__/division.py
@@ -1,0 +1,26 @@
+from __future__ import division
+
+def _check(x, test):
+    # Floats have different precision/representation in js and python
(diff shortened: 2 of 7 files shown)


all3fox:
I also realize that this is kind of a gimmick: I am not testing the empty `__future__` model, I am just making sure that a) the thing is importable and b) it kind of works as expected, even though it does not come from `__future__`

JdeH:
Your test tests exactly what should be tested, as far as I can see.

all3fox:
Any idea why the `with_statement` does not work?

JdeH:
I've made them into a regular autotest (back to back comparison) by adding a dummy sys module and a few lines of code to your autotest.py, since newer CPython versions do not look for modules in the current directory.

all3fox:
Yes, sure, but I do not understand what for. Are you saying that without those lines the following did not work:

```
cd transcrypt/development/continuous_integration
./test_server.py 8080
```

Then navigate to `http://127.0.0.1/chk/__future__` and see all the tests pass.

Do you do something else to run the tests (and need the `sys` module for that)?

JdeH:
It's just that all other autotests have some visual output in the browser.

I've used the sys module in your file autotest.py as follows:

```
try:
    sys.path.append ('.')
except:
    pass

```

This is needed in CPython 3.6 because otherwise autotest.py won't find its imports.
This is a recent change in CPython: the current dir isn't in the module search path anymore.

To be able to compile the same code in Transcrypt and CPython it was put in a try clause.
There's no point in putting the import in a try clause as well, since importing is done compile time by Transcrypt.

PR #348, opened 2017-06-25
Add inspect.isclass and tests for it

all3fox:
Currently the javascript version returns `False` for everything (class, function or whatever) while python version, obviously, returns `False` for everything except a class, when in returns `True`. See test.

Any idea why `isclass` does not work?

Update: I copied its implementation [from cpython](https://github.com/python/cpython/blob/af5392f5c6f8014659e995840df6ee7b5017f743/Lib/inspect.py#L73).

### DIFF
--- a/transcrypt/development/automated_tests/inspect/autotest.py
+++ b/transcrypt/development/automated_tests/inspect/autotest.py
@@ -1,0 +1,9 @@
+from org.transcrypt.autotester import AutoTester
+
+import test_isclass
+
+atester = AutoTester()
+
+atester.run(test_isclass, 'test_isclass')
+
+atester.done()
--- a/transcrypt/development/automated_tests/inspect/test_isclass.py
+++ b/transcrypt/development/automated_tests/inspect/test_isclass.py
@@ -1,0 +1,20 @@
+import inspect
+
+class SimpleClass:
+    pass
+
+def simple_function():
+    pass
+
+def run(test):
+    test.check(inspect.isclass(42))
+    test.check(inspect.isclass(42.0))
+    test.check(inspect.isclass('foo'))
+
+    simple_class = SimpleClass()
+    test.check(inspect.isclass(SimpleClass))
+    test.check(inspect.isclass(simple_class))
+    test.check(inspect.isclass(SimpleClass()))
+
+    test.check(inspect.isclass(simple_function))
+    test.check(inspect.isclass(simple_function()))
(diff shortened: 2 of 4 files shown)


the1337guy:
Just a quick note, don't name the argument 'object', it's already a standard builtin.

the1337guy:
Even tho I have no idea why CPython names it a object.

all3fox:
Ok, so I see an implementation of [`isinstance` in transcrypt][1] is here. This probably causes the mistake: `isinstance` assumes two arguments, an `object` and a `classinfo` and I should probably pass something into the `classinfo` field that is not `type`. Dunno what though.

[1]: https://github.com/QQuick/Transcrypt/blob/2c623f6dec2da49ee6b523e6ae376ed1650ce031/transcrypt/modules/org/transcrypt/__javascript__/__builtin__.mod.js#L294

all3fox:
Ok, the penny partially dropped: Transcrypt writes its own builtins and writes its own `isinstance`. So, if I rely on the convention that all classes have the `__metaclass__` attribute, I'm nearly there.

The only hiccup is in the updated tests: once construction happens and there is an instance of a class, the `__metaclass__` attribute is still there (which makes current `isclass` say that an instance of a class is still a class). This is not how python does it: an instance of a class is no longer a class:

```
x = SomeClass()
inspect.isclass(x) # False because isinstance(x, type) is False
isinstance(x, SomeClass) # True, because it is a class instance
```

Any thoughts how this could be fixed? Is removing the `__metaclass__` attribute from a constructed class too much of a breakage? Any other tricks/hacks to imitate `inspect.isclass` that I am missing?

all3fox:
Oh, I've messed up the branch (used the currently unmerged `bottle` changes), please wait till I untangle it (the single commit will appear below this message and I will write an update).

JdeH:
Objects are created by the following line in __core__.mod.js (156):

`var instance = Object.create (this, {__class__: {value: this, enumerable: true}});`

From this statement it appears that objects have a __class__ attribute, while classes don't.
This should be enough to make the distinction.

Disclaimer: It's quite a while ago I wrote this, so I may have overlooked something. ;)

all3fox:
Thanks for the help! I was going to add more functionality (`inspect.ismodule` and `inspect.isfunction` are good candidates) but I am generally back to "reading" Transcrypt source code now.

This could be merged as is, thanks!

JdeH:
Good thing that you also provided an autotest / ci test!

JdeH:
In addition I've also expanded the shipment test.

```
        test ('development/automated_tests/hello', 'autotest', True, False, switches)
        test ('development/automated_tests/transcrypt', 'autotest', True, False, switches + '-c ')  
        test ('development/automated_tests/time', 'autotest', True, False, switches)    
        test ('development/automated_tests/re', 'autotest', True, False, switches)    
        test ('development/automated_tests/inspect', 'autotest', True, False, switches)  
```

It will be in the next commit.

PR #350, opened 2017-06-28
Asteroids demo in 3d

theodox:
This is a simple version of Asteroids written in Python using `three.js` for rendering. It also includes a complete API wrap of `three,`, although only a small subset of the functionality is used.   Some other things worth noting:

* Use of generators for timed wait coroutines
* use of HMTL5 audio components
* basic class inheritance
* properties
* `nonlocal` scope
* python functions used as JS callbacks

copyright credits for the audio are included in the game over screen


JdeH:
Wow!!!

JdeH:
Hi, finally found the time to play with this.
Really works well!
Find it hard to control spaceship speed & direction...
I must say I haven't got much gaming experience.
I'll try to make this avaiable online through the gallery.
Thanks a lot!

theodox:
If I get a bit of time I'll change the spaceship to a cone, it makes it
much clearer.  The lag on the controls is my effort to recreate the
original, which had a lot of inertia on the controls.  it's tweakable in
the control handlers though one wants to make it more responsive

JdeH:
Indeed seeing which side is front will help.

PR #351, opened 2017-06-29
[WIP] Rewrite __main__.py to use logging and yaml config

all3fox:
Sorry, I suddenly find out that I am very much out of time to work directly on Transcrypt, so I will probably not finish this PR. Anyway, it shows the main directions I wanted to suggest changing:

1. Bring the parse from the deep `modules/org/transcrypt/utils.py` up front to `__main__.py`
2. Rename `__main__.py` to something more human-friendly (but this might not be what you want, as I [mention on gitter][1]).
3. Use builtin `logging` (because it can do loglevels, log exceptions, be flexibly configured, etc) instead of the in-house `def log():` and the need to be printing newlines/tracebacks explicitly
4. Use external libraries through `requirements.txt` (like `pyaml` to configure logging in two lines)
5. Separate configuration constants into a dedicated `transcrypt.yml` file instead of all over the code.

* Simplify error reporting (no custom exit codes)
* Do not subclass a parser, just set it up better
* Use logging for granular levels/exceptions reporting
* Use pyaml (via pip install -r requirements.txt) to configure

Unfinished, does not work yet. [Related gitter discussion.][1]

[1]: https://gitter.im/Transcrypt/Lobby?at=5954adff11755ab05633ff38


all3fox:
Oh, I find that many packages actually have `__main__.py` and an entry point of `def main`, like here:

https://github.com/jakubroztocil/httpie/blob/f1d4861faed6486a356175ce9f92d4da96ddba01/setup.py#L84

PR #353, opened 2017-07-04
added on-screen instructions

theodox:
* Ship is now a proper cone (thanks to up…dated three.js link); 
* reset cursor at game end
* format fixes


JdeH:
Should make playing more fun...

PR #359, opened 2017-07-11
Add some of the string methods:

all3fox:
- isalnum
 - isalpha
 - isdecimal
 - isdigit
 - islower
 - isupper
 - isspace
 - isnumeric (*)

These methods are *not* entirely compatible with CPython because they
deal with ascii characters only, not general unicode. Also, isnumeric
existed before and it is even less compatible: it allows floats, which
python does not.

I have also tried adding some tests but it did not work for me:

I navigated to `automated_tests/transcrypt`, compiled the tests with these two commands:

```
transcrypt -c -da autotest.py
transcrypt -r -c -da autotest.py
```

After that, I opened `autotest.html` and it showed me an empty table.


JdeH:
Great!
I'll try to fix the test some time later.

JdeH:
Will be a while.

PR #362, opened 2017-07-17
Expand projectDir and suggest to drop programDir

all3fox:
For me `Transcrypt` is often run as part of a script that makes `Transcrypt` transpile files that are deeper than one level down in the directory, like so:

```
./bin/transcrypt ./dir0/dir1/file.py
```

Yet `__main__.py` relies on `sys.path.getcwd()` for `programDir` which is later propagated further and further until it is used as the directory where the file-to-be-transpiled (aka `file.py`) resides. The use-case above does not always work because of that. Instead, I would like to propose the reuse of `utils.commandArgs.source` which already points to the file that must be transpiled. I would also like to propose dropping `programDir`.

### DIFF
--- a/transcrypt/__main__.py
+++ b/transcrypt/__main__.py
@@ -102,7 +102,7 @@ def main ():
         # Prepend paths that are needed by transpiled or executed user code, since they have to be searched first
         # So user code favors Transcrypt modules over CPython modules
         extraPath = utils.commandArgs.xpath.replace ('#', ' ') .split ('$') if utils.commandArgs.xpath else []
-        projectPath = [programDir] + extraPath
+        projectPath = [programDir, utils.commandArgs.source.rsplit('/', 1)[0]] + extraPath
         
         sys.path [0 : 0] = projectPath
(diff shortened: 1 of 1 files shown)


JdeH:
This is actually a very good suggestion.

Only there are some complications:

- Windows uses '\\' rather than '/', this requires a slight mod of your code
- There may not always be a path, since (under Windows) the first command line arg may simply be a filename, without even a reference to the current dir

However these can easily be overcome, and I have already adapted your code (so no need to do that).
Moreover I have not been consistent in my use of the terms path and dir in __main__.py, which I like to correct on this occasion.

I will try to adapt it accordingly.
So while in its current form I will not merge your pull request, it is an important improvement and the idea behind it will be fully conserved.

Please let me know my version works for your indicated usecase as soon as I committed it.

Kind regards
Jacques

JdeH:
Committed as 3.6.45

all3fox:
Yes, thanks, now works as expected for me

PR #372, opened 2017-08-10
Fix the py_update for sets: break down each iterable into components

all3fox:
Currently, if you do something like:

```
s = set()
s.update('hello')
```

then transcrypt will translate this into a set object that holds one string `'hello'`, while python breaks it down into letters in alphabetical order.

This PR fixes this inconsistent behavior and adds a couple of tests to cement it.

Update: also apologies for removing a lot of extra whitespace in the files that I've touched. My editor does whitespace cleanup automatically and turning it off for this PR proved to be hard so I gave up.

Update2: nice, I see you've fixed the CI tests! That's good to hear, well done!

### DIFF
--- a/transcrypt/development/automated_tests/transcrypt/set_comprehensions/__init__.py
+++ b/transcrypt/development/automated_tests/transcrypt/set_comprehensions/__init__.py
@@ -1,46 +1,70 @@
 def run (autoTester):
     even = {2 * i for i in [0, 9, 1, 7, 2, 8, 3, 6, 4, 5]}
     autoTester.check (even)
-    
+
     odd = {2 * i + 1 for i in [5, 6, 7, 8, 9, 4, 3, 1, 2, 0]}
     autoTester.check (odd)
-    
+
     even.add (12)
     even.add (12)
     autoTester.check (even)
-    
+
     even.discard (12)
     even.discard (12)
     autoTester.check (even)
-    
+
     uni = even.union (odd)
     autoTester.check (uni)
-    
+
     autoTester.check (odd.isdisjoint (even))
     autoTester.check (uni.isdisjoint (even))
-        
+
     autoTester.check (even.issuperset (uni))
     autoTester.check (uni.issuperset (even))
-    
+
     autoTester.check (even.issubset (uni))
     autoTester.check (uni.issubset (even))
-    
+
     first = {4, 1, 0, 5, 3, 2, 6}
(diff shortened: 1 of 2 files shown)


JdeH:
It's a bit too much work for me to repair whitespace manually...
But I've again taken the core of what you propose, did some slight optimization, and commited it.
The tests I've put in the data_structures testlet, since the list_comprehensions testlet is only about those.

    Array.prototype.py_update = function () {   // O (n log n)
        var that = this.slice ();   // Make a shallow copy of existing set
        for (var i = 0; i < arguments.length; i++) {
            var argument = arguments [i];
            if (Array.isArray (argument)) {
                that = that.concat (argument);
            }
            else {                  // Needed for strings
                for (var j = 0; j  < argument.length; j++) {
                    that.push (argument [j]);
                }
            }
        }
        
        that.sort ()                // Put identical elements next to each other

        this.clear ();              // Clear existing set to prevent dupplicates
        for (var i = 0; i < that.length; i++) {
            if (that [i] != that [i - 1]) {
                this.push (that [i]);
            }
        }
    };

all3fox:
Sure, looks good, thanks!

A few questions though:

1. Your for loops increment the counter by one, yet the check is `<`. Why are you avoiding an identity check? I've always believed that identity check is (probably negligibly) faster than inequality.
1. You use "equality" operators rather than "identity" operators (i.e. `==` rather than `===`). Again, "identity" operator is never slower than "equality" and many js linters would recommend using "identity" whenever possible. Is there some reasoning behind using equality?

JdeH:
1. It's a general habit, picked up the 1970's from K&R C. The stepsize may differ from 1, in which case you never exactly hit the target. Also the loop"counter" may be a float, which may miss the target because of rounding. I try to have fixed, robust habits in coding.

2. Once again a multi-language thing (as is my camel case var naming style). All languages have ==, only JS has ===. Unless there's a compelling reason I prefer the least common denominator idiom, to be able to switch & combine languages in the simplest fashion.

I don't expect anyone to agree with these habits... When you've been programming for 45 years, you tend to acquire certain ways. No value judgement meant here.

Your code was fine, everyone has to cut his/her own path through the language jungle...

all3fox:
Hey, thanks for the write-up! Anyway, I am working on my next PR and notice that the changes for `py_update` are not on master yet. Please push them upstream when you have time, otherwise the history will become a little hard for you to untangle.

PR #373, opened 2017-08-11
Add additional options so autotest.html will load without errors

tubaman:
When transpiling and loading autotest.html as suggested in the
[documentation](http://www.transcrypt.org/docs/html/installation_use.html)
an `Uncaught ReferenceError: complex is not defined` error is thrown.
Add additional command line options so that the page will load without
error.


JdeH:
Good improvement, thanks!
Will regenerate the online docs.

PR #376, opened 2017-08-14
Make __t__ respect __bool__ and __len__ like Python does

### LINKED ISSUE #375: Custom classes with `__bool__` or `__len__` have wrong `__t__` results
Even with the `-t` switch for overloaded truth-value testing, there are problems for custom python classes.

Let's begin with a simple and empty class of:

```
class Foo:
    pass
```

This class defines neither `__bool__`, nor `__len__`, so instances of this class are true in python by default. Transcrypt will say this class is true because `__t__` will check `foo instanceof Function` which is always the case: Transcrypt uses custom `__class__` function to define classes, which in turn returns a function.

Now, if you start defining your class further with `__bool__` or `__len__`, Transcrypt will not get to calling these methods at all. Even if we forget for a second about the `instanceof Function` which is always true, the next step in `__t__` is calling `len(foo)`, which is again a custom function that will try the `.length` attribute and will also try to count the "number of non-special attributes" as determined by custom function `__specialattrib__`.

The `__len__` and `__bool__` methods of the object itself are never called, am I correct? If that is the case, what is the reasoning for it?

I will be fooling around with `__t__` and `len`, see how the existing tests react to that and submit a PR.

all3fox:
This improves compatibility and allows more fine-grained truth-value
testing control for python classes.

Closes #375 

There are more tests and the old tests pass locally, let's look at what CI has to say.

Update: and CI passes, bingo!

### DIFF
--- a/transcrypt/development/automated_tests/transcrypt/truthyness/__init__.py
+++ b/transcrypt/development/automated_tests/transcrypt/truthyness/__init__.py
@@ -112,5 +112,28 @@ def run (autoTester):
     class A:
         pass
         
+    class B:
+        def __bool__(self):
+            return False
+
+    class C:
+        def __bool__(self):
+            return True
+
+        def __len__(self):
+            return 0
+
+    class D:
+        def __len__(self):
+            return 0
+
+    class E:
+        def __len__(self):
+            return 1
+
+    autoTester.check ('instances of a custom class')
     autoTester.check (not not A ())
-    
+    autoTester.check (not not B ())
+    autoTester.check (not not C ())
+    autoTester.check (not not D ())
+    autoTester.check (not not E ())
--- a/transcrypt/development/automated_tests/transcrypt/truthyness/__init__.py
+++ b/transcrypt/development/automated_tests/transcrypt/truthyness/__init__.py
@@ -113,27 +113,28 @@ def run (autoTester):
         pass
         
     class B:
-        def __bool__(self):
+        def __bool__ ( self):
             return False
 
     class C:
-        def __bool__(self):
+        def __bool__ (self):
             return True
 
-        def __len__(self):
+        def __len__ (self):
             return 0
 
     class D:
-        def __len__(self):
+        def __len__ (self):
             return 0
 
     class E:
-        def __len__(self):
+        def __len__ (self):
             return 1
 
-    autoTester.check ('instances of a custom class')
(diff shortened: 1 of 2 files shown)


all3fox:
Turns out I missed the `undefined` and the `null` cases yesterday. It is also not clear to me how to add those cases to the tests (not sure if that's possible at all), but I've fixed my code for those two special cases.

JdeH:
I am in doubt.

Your tests indicate this functions correctly, but it is quite a lot of code and if the library grows by the same factor on many cases, Transcryp will be far less compact, to achieve a bit more compatibility. I still believe in the 80/20 rule.

But on the other hand, this may be on the right side of the line, since it is quite fundamental. 

What I would like is to achieve the same effect with a more compact (and maybe faster) conditional expression. The original conditional expression is already quite complicated and took a lot of thought. I am not sure incorporating this new behaviour in it is feasible. Maybe it is.

So indeed, still in doubt...
Is this acually used in your "boolean" code? Probably, given the name...

JdeH:
Have edited your code to a conditional expression.
Still, while your code functions correctly, the conditional expression turns out to give erroneous results for now.

It is:

```
    function __t__ (target) { 
        return (
            // Avoid invalid checks
            target === undefined || target === null ? target :
            
            // Take a quick shortcut if target is a simple type
            ['boolean', 'number'] .indexOf (typeof target >= 0) ? target :
            
            // Use __bool__ (if present) to decide if target is true
            target.__bool__ instanceof Function ? (target.__bool__ () ? target : false) :
            
            // There is no __bool__, use __len__ (if present) instead
            target.__len__ instanceof Function ?  (target.__len__ () !== 0 ? target : false) :
            
            // There is no __bool__ and no __len__, declare Functions true.
            // Python objects are transpiled into instances of Function and if
            // there is no __bool__ or __len__, the object in Python is true.
            target instanceof Function ? target :
            
            // Target is something else, compute its len to decide
            len (target) !== 0 ? target :
            
            // When all else fails, declare target as false
            false
        );
    }

```

JdeH:
Seems I've deleted your initial comment, will add it in next commit.

JdeH:
Google closure does a good job at times.
About compatibility: Your code (and the resulting c.e.) were in fact quit compact.
Still it remains a hard decision each time.
If everything is incorporated it will be more like Brython, which certainly earns its place under the sun, but, as I see it, in a different niche.

PR #378, opened 2017-08-18
Make function len call __len__ (from Python) if it exists

all3fox:
I've also added a couple of tests for the `len` function. Notably, it is still rather different from the `len` in Python: it does not throw errors for things that "do not have a length", like numbers, `None` or whatever. However, such fine-grained compatibility is not necessary. What is necessary, in my opinion, is to use `__len__` when the object has it.


JdeH:
Looks fine to me!

all3fox:
Wow, thanks, this is some impressive speed :+1:

If you can remember of the top off your head, I would appreciate it if you explained the "special attributes" case: why would you need to count the "non-special" attributes of an object and return it as its length?

Issue #394, opened 2017-10-26
ValueErrors cannot be constructed with a message in Transcrypt


daboross:
In Python 3.6, the following code will produce an exception with a message:
```
raise ValueError('this specific error occurred')
```

```
>>> raise ValueError('this specific error occurred')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: this specific error occurred
```

In Transcrypt, the same code generates a generic "Erroneous Value" error message.

I believe the same is true of "KeyError". I'll try to send a PR for this if this is something we can change - I don't know what code relies on "Erroneous Value" being the default message in Transcrypt - this is opposed to the default message of an empty message in Python 3.6.

daboross:
Looking back, I see this was introduce in https://github.com/daboross/Transcrypt/commit/9924f308b3711602ac282f1761c38d3487f3b84a#diff-493186bca6f730b8f74b3c21e15ab988.

I can't seem to find any significance of `"Erroneous Value"` and `"Invalid Key"`, but I would definitely like to know if there is some.

JdeH:
Solution was merged.
Thanks.
Feel free to reopen if I overlooked anything!

PR #395, opened 2017-10-26
Remove default messages for ValueError and KeyError

daboross:
I believe this better matches Python's default behavior. 

This commit removes default `"Erroneous Value"` and `"Invalid Key"` messages for `ValueError` and `KeyError` respectively. These do not match Python's behavior, and prevent more custom messages from being passed in.

This commit also updates usages of ValueError and KeyError in `__builtin__` to use individual strings which match python3.5's error messages rather than using the defaults that are now removed.

I don't know if any end-user code relies on these messages, but this could be changed to keep `"Erroneous Value"` and `"Invalid Key"` messages as default if none are supplied. I chose to have `""` as the default message in order to match Python 3.5's behavior, but that's not as essentially useful as allowing customizable messages in the first place.

I see the messages were introduced in https://github.com/QQuick/Transcrypt/commit/9924f308b3711602ac282f1761c38d3487f3b84a#diff-493186bca6f730b8f74b3c21e15ab988, but I don't know the significance of `"Erroneous Value"` and `"Invalid Key"` here. If these messages do carry weight, I would definitely like to know.

The new code for `ValueError` and `KeyError` accepts an optional `'message'`, similar to what Exception itself accepts. The default string representation for a new `ValueError()` is now `""` (to match `ValueError()`'s message in Python3.5).

See #394.

### DIFF
--- a/transcrypt/modules/org/transcrypt/__javascript__/__builtin__.mod.js
+++ b/transcrypt/modules/org/transcrypt/__javascript__/__builtin__.mod.js
@@ -286,7 +286,7 @@ __pragma__ ('endif')
                 return 1;
             }
             else {  // Needed e.g. in autoTester.check, so "return any ? true : false" won't do
-                throw ValueError (new Error ());
+                throw ValueError ("could not convert string to float: '" + str(any) + "'", new Error ());
             }
         }
         else {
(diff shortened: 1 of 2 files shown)


JdeH:
Getting closer to CPython behavior at no cost is a good thing.
Hadn't thought too carefully about this!

PR #408, opened 2017-11-26
Translate `import x.y.z` statements at compile time.

daboross:
Instead of using `__nest__` at runtime, produce nested objects at compile time.

This generates `var x = {'y': {'z': ...}};` rather than `var x = {}; __nest__(x, 'y.z', ...);` for `import x.y.z`:

```python
import a.b.c
```
changes:
```diff
-var a = {};
-__nest__ (a, 'b.c', __init__ (__world__.a.b.c));
+var a = {'b': {'c': __init__ (__world__.a.b.c)}};
```
```python
import a
```
changes:
```diff
-var a = {};
-__nest__ (a, '', __init__ (__world__.a));
+var a = __init__ (__world__.a);
```

The result is faster runtime evaluation of these imports: we've already parsed 'x.y.z', and instead of copying all attributes of 'z' to a new object, we just reference the existing 'z' module object.

This does have two side effect changes:
- Changing imported modules's values changes now that module's exported values
  - Previously, it only changed a copy of the module that was unique to each module importing it
- Importing within a function no longer generates a module-level variable
  - I believe https://github.com/QQuick/Transcrypt/pull/271 made a similar change to this for `from x import y`, but `import x(.y.z)` statements still currently have a module-level variable when used within a function?

### DIFF
--- a/transcrypt/modules/org/transcrypt/compiler.py
+++ b/transcrypt/modules/org/transcrypt/compiler.py
@@ -492,7 +492,6 @@ class Generator (ast.NodeVisitor):
         self.classDecorators = []
         self.use = set ()
         self.all = set ()
-        self.importHeads = set ()
         self.docString = None
         self.docStringEmitted = False
         self.lineNr = 1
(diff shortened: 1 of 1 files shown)


daboross:
This may be broken.. trying to figure out what exactly is failing and if I can fix it now.

daboross:
Nevermind, I see what's broken about this. This breaks `import x; import x.y;` if both are used.

PR #411, opened 2017-11-30
Refactored the pragma if statements into pragma.py with a registry.

doconix:
The point of this change is to allow "plugin" pragmas.  This would allow any project using transcrypt to add additional pragmas without requiring changes to the transcrypt codebase.  Before, the code had a big `if/elif/elif/...` sequence to determine which pragma to handle.

This new change instead allows pragma functions to register within a "registry" (just a `dict`).  When `__pragma__` is encountered, the appropriate handler function is retrieved from the dict and run.  At python startup, the various handler functions register with something like `@pragma_handler('nokwargs')`.

If a project using transcrypt needs additional pragma handlers, it simply needs to import the `pragma_handler` decorator and use it on any function.

Two important notes if this is accepted:
* Someone should also refactor the pragmas that are handled elsewhere in `compile.py` to follow this same pattern. I didn't do this because I'm not sure I understand the other code yet and the reasons they were separated.
* I have run the autotests successfully, but someone needs to do more testing to ensure I didn't break any of the pragmas. I'm not familiar enough with the project to know that I ran all the tests correctly.

### DIFF
--- a/transcrypt/modules/org/transcrypt/compiler.py
+++ b/transcrypt/modules/org/transcrypt/compiler.py
@@ -24,10 +24,9 @@ import datetime
 import math
 import traceback
 import io
-import subprocess
-import shlex
 
 from org.transcrypt import __base__, utils, sourcemaps, minify, static_check, type_check
+from org.transcrypt.pragma import lookup_pragma_handler
 
 class ModuleMetadata:
     def __init__ (self, program, name):
@@ -1449,22 +1448,6 @@ class Generator (ast.NodeVisitor):
 
             self.emit (')')
             
-        def include (fileName):
-            searchedIncludePaths = []
-            for searchDir in self.module.program.moduleSearchDirs:
-                filePath = '{}/{}'.format (searchDir, fileName)
-                if os.path.isfile (filePath):
-                    return tokenize.open (filePath) .read ()
-                else:
-                    searchedIncludePaths.append (filePath)
-            else:
-                raise utils.Error (
-                    lineNr = self.lineNr,
-                    message = '\n\tAttempt to include file: {}\n\tCan\'t find any of:\n\t\t{}\n'.format (
-                        node.args [0], '\n\t\t'. join (searchedIncludePaths)
-                    )
-                )
-
         if type (node.func) == ast.Name:
             if node.func.id == 'type':
                 self.emit ('py_typeof (')   # Don't use general alias, since this is the type operator, not the type metaclass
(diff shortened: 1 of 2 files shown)


JdeH:
I've been giving this some thought, and I am not (yet) convinced.

At first thought I can see two advantages of your proposal:

1. Separation of concerns (modularisation) and having thin interfaces (loose coupling) is generally a good matter. Your proposal attempts to separate the __pragma__ mechanism from the rest of the code, which seems to lead to less spurious dependencies (less "spaghetti").

2. Empowerment of the user. Transcrypt can't be everything to everyone. By enabling the user / developer to roll her own pragma's, we would allow thousand flowers to grow. The Transcrypt team could never provide solutions for anything any user of it may ever encounter. Giving the user as much customization power in her hands as possible is the only reasonable answer.

While you mainly elaborate 2 in your motivation, I consider 1 at least as important.

BUT...

To start with 1:

If I look carefully at the existing pragma's I notice that many of them require some kind of special, deeply engrained support in the code in compiler.py.

Looking e.g. at the following bunch:

```
               if node.args [0] .s == 'alias':		 +                func = lookup_pragma_handler(node.args [0] .s)
 -                    self.aliasers.insert (0, self.getAliaser (node.args [1] .s, node.args [2].s))		 +                if func is not None:
 -                elif node.args [0] .s == 'noalias':		 +                    func(self, node)
 -                    if len (node.args) == 1:		
 -                        self.aliasers = []		
 -                    else:		
 -                        for index in reversed (range (len (self.aliasers))):		
 -                            if self.aliasers [index][0] == node.args [1] .s:		
 -                                self.aliasers.pop (index)		
 -		
 -                elif node.args [0] .s == 'noanno':		
 -                    self.allowDebugMap = False		
 -		
 -                elif node.args [0] .s == 'fcall':		
 -                    self.allowMemoizeCalls = True		
 -                elif node.args [0] .s == 'nofcall':		
 -                    self.allowMemoizeCalls = False		
 -		
 -                elif node.args [0] .s == 'docat':		
 -                    self.allowDocAttribs = True		
 -                elif node.args [0] .s == 'nodocat':		
 -                    self.allowDocAttribs = False		
 -		
 -                elif node.args [0] .s == 'iconv':       # Automatic conversion to iterable supported		
 -                    self.allowConversionToIterable = True		
 -                elif node.args [0] .s == 'noiconv':     # Automatic conversion to iterable not supported		
 -                    self.allowConversionToIterable = False		
 -		
 -                elif node.args [0] .s == 'jsiter':      # Translate for ... in ... : ... literally to for (... in ...) {...},		
 -                    self.allowJavaScriptIter = True     # to enable iterating JavaScript objects that are not dicts		
 -                elif node.args [0] .s == 'nojsiter':    # Dictionary keys without quotes are identifiers		
 -                    self.allowJavaScriptIter = False		
 -		
 -                elif node.args [0] .s == 'jskeys':      # Dictionary keys without quotes are string literals		
 -                    self.allowJavaScriptKeys = True		
 -                elif node.args [0] .s == 'nojskeys':    # Dictionary keys without quotes are identifiers		
 -                    self.allowJavaScriptKeys = False		
 -		
 -                elif node.args [0] .s == 'jsmod':       # % has JavaScript behaviour		
 -                    self.allowJavaScriptMod = True		
 -                elif node.args [0] .s == 'nojsmod':     # % has Python behaviour		
 -                    self.allowJavaScriptMod = False		
 -		
 -                elif node.args [0] .s == 'gsend':       # Replace send by next.value		
 -                    self.replaceSend = True		
 -                elif node.args [0] .s == 'nogsend':     # Don't replace send by next.value		
 -                    self.replaceSend = False

```

all of these pragma's set or reset some kind of switch that triggers a whole lot of behaviour that is distributed through many places of the compiler. Just being able to add the code pertaining to a section in the big if..elif ...elif...else... wouldn't help much in really creating the functionality for any of these pragma's, they require deep modification of compiler.py anyhow.

The `js` and `xtrans` pragma's are remarkable exceptions to this rule: Since they just interject code that has very little context depencency they indeed may be factored out as self contained functions looked up in a dictionary.

So while having a separate registry of pragma's _suggests_ separation of concerns by insultating pragma's syntactically, they are still semantically tied to compiler.py by many strands of spaghetti. It's the nature of the animal, I am afraid: Most pragma's intervene deeply in a diverse range of activities in the compilation process. That's their power (flexibility) and their weakness (complexity).

Indeed some pragma's aren't in the big _if_. For each of those there were pressing reasons to put them somewhere else, as indicated by the commentin the following fragment of compiler.py:

if node.args [0] .s in ('skip', 'noskip', 'ifdef', 'ifndef', 'else', 'endif'):		
 -                    pass                                # Easier dealth with on statement / expression level in self.visit

I'd rather have them all in one place, but the complexities outweighted the benefits in my experience.

Lets look at 2:

Could a user of Transcrypt provide her own pragma's. As far as my insight goes now (but I may overlook something): not easily to say the least. Defining the functions and adding them to the "registry" would have to happen in code _translated by Transcrypt_, rather than in the code of Transcrypt itself. Crossing the border between compiler and compilee isn't trivial as can be seen in the code of e.g. the `_js_` pragma, expecially the _\_\_include\_\__ parameter. I can't currently see that it would be possible in any relevant cases, given the deep semantical ties between the code in _compiler.py_ and most pragma's.

Moreover:

Yet another problem is that users are tempted to make their pragma's deeply dependent upon the guts of compiler.py. This means that with each changes of these guts, we would destroy the work of a number of users. Since we don't have their code as part of Transcrypt in that case, we can not foresee or prevent that. And even if we knew what they depend on, we can't allow compiler.py to be effectively functionally frozen.

So from a future perspective of continued improvement of Transcrypt I'd say this isn't desirable, since at each change we would antagonize some users having written their own pragma's. It should be clear to them that their investment in code is safe. With user defined pragma's that would generally not be the case, unless they stand very separate from the rest of the code. But as said, that's rare.

So:

All in all, while I consider it a bright idea worth looking at carefully, currently I'd say it has some fundamental problems and a limited yield. I think it would be wise not to follow this path.

But of course I may have missed some essential points. So I am curious about your reaction and open to be convinced. Rightnow I'm not (convinced, that is...).

Kind regards
Jacques

doconix:
Thank you for the thoughtful response. I appreciate you taking a serious look at the PR.  I agree with your judgement that the pragma switches are intricately tied to the compiler. I thought through it when I was starting to make the code changes, but since most pragmas just set switches, I didn't consider the ties as strong as you explain here.  Switches are set (usually through get/set) from outside objects all the time.  But I see in your response how these are not just "regular" switches -- allowing users to mess with them could seriously screw up the compiler operation.

You are right that my perspective is mainly being driven from the `xtrans` one, which is not typical.  The PR might overreach the use case I am after.  

Here's the use case I was after: right now, the only way to declare "special" code blocks that transcrypt should leave alone (and instead give to me to handle) is with pragmas: specifically the `js` and `xtrans` ones.  The `js` one is limited because it only says "leave this alone" rather than "let me process it".  The `xtrans` is also limited because it requires me to run a shell program.  Letting the user define additional pragmas is a way to allow the user to do anything to the block, such as call a python function.

So my idea is not really to allow the user to modify the compiler process but to "format" or "process" or "take control of" a specific block of code during the transpile process.  The plugin registry is just a nice way to tell the compiler when to hand things to me.

An alternative that still meets this use case is to create a single pragma word (or use the new `xtrans` word) and put the entire plugin system below that single pragma.  

Another alternative is to create a different system than `__pragma__` that does this.  Something like `__xtrans__` that allows me to plug in things.

In summary, I agree with your concerns.  I just considered the ability to extend the system more important.  But in light of this discussion, and if you want to get this functionality supported, I would recommend my first alternative.

I need to state here at the end of my comment that your addition of the current `xtrans` pragma **already solved my immediate need**. My PR was just a way to give back a bit.  So you are welcome to drop this entirely if you don't want to mess with it.  Or you can ask me to do one of the alternatives above and I can submit another PR. I'm good with any of it.

JdeH:
Hi again,

Suppose adding user defined pragma handlers would be possible in some limited form, how then would you supply the Python functions to handle them, even if they were merely processing some code without much dependencies upon compiler.py.

The user's code is currently _translated by_ Trancrypt rather than _used by_ Transcrypt.

Seems I miss an essential part of what you propose...

Adding limited possibilities while protecting the user from depending too much upon the innards of Transcrypt may be a useful addition. Do you propose to use "exec" or "eval" on functions that are somehow present in the code to be compiled, may suppied as a string to the user defined pragma?

Maybe a simple example worked out in some detail would help in clarifying what exactly you have in mind.

doconix:
I get what you are saying about Transcrypt not "running" the python.  The functions would have to be in a different file somewhere.  As long as the python path includes the current directory, the user's module could be in or below that directory.

The following is along the lines of what I'm thinking:

```
__pragma__ ('xfunc', 'org.me.mymodule.makeupper', '{}', 'code/str/etc here')
```

When Transcrypt sees this pragma, it loads `mymodule` and executes the function.  This code gets a ref to a function inside a dynamic module:

```
def import_by_name(name):
    mod_name, obj_name = name.rsplit('.', 1)
    mod = importlib.import_module(mod_name)
    return getattr(mod, obj_name)
```

The user function in `org/me/mymodule.py` might look like:

```
def makeupper(st):
    return st.upper()
```

Note 1: I am trying to decide whether a single pragma is best for the above or a start/stop pragma set similar to `fcall` and `nofcall` with the code in between.  But that's just an implementation issue.

Note 2: The word `xfunc` is no good. If you add this functionality, we need a better word. :)

Note 3: This could also use the registry idea, but I'm not sure it's worth the extra complexity.

Note 4: I don't believe there is a security issue in having Transcrypt call a dynamic function because this all happens at build time, not at runtime.  The end users can't specify the function, for example, because the transpile is done by the time the web server runs.

Finally, I want to stress again that I'm not hung up on any of this.  It's your project, so do whatever you think is best.  I'm OK if you don't want to add this to the project.

Best, Conan

JdeH:
A start-stop pragma pair would have to intercept the code between the start and the stop, which isn't trivial.
So a simple pragma seems, well, simplest.

doconix:
That's fine.  Just let me know how to proceed.  If you want me to try another PR with these new ideas,  I can do so.  If you want to put this on hold for now, that's fine too.  I'm happy to help but leave the decision entirely up to you.

Cheers.  I see you are Dutch.  I have been to Maastricht in the south of Netherlands.   Beautiful.  Absolutely beautiful.  I loved the week I spent in NL. Although I got picked up by a traffic camera and got a ticket.  Here in the Western USA, traffic cameras that give speeding tickets wouldn't last long because they'd end up with a .22 bullet in them.  Still a bit of the wild west .  Ha.

JdeH:
We sometimes have them too, traffic camera's with bullet holes in them. Only that's rare now, since so many spots have multiple surveillance camera's that the license plates of the shooter would be picked up somewhere.

I've been to the USA too, though not often. Several places in Texas (work), SanFrancisco (work), New York (leisure) with my son. Didn't know skyscrapers can be beautiful. But some of them actually are. Liked Central Park best, though. Such peace and quiet in the middle of the town...

If I would be more of a traveler I'd probably attempt to hold a conference talk about Transcrypt somewhere, but I am not.

About the PR:

Suppose indeed there were a pragma along th lines you propose:

> __pragma__ ('xfunc', 'org.me.mymodule.makeupper', '{}', 'code/str/etc here')

what would be a practical usecase for it?

I mean, what type of rather context independent code transformation should it supply?
Capitalizing text doesn't seem very useful.
It's not much work making this mechanism and security isn't a concern indeed, but I am struggling to think of a case where such a mechanism could be of help.
Can you supply me with a practical example of this.

JdeH:
This currently seems to have come to a standstill.
I'll close for now but feel free to reopen.

PR #435, opened 2018-01-06
datetime module

neriusmika:
More testing needed. If you found some errors, I will fix them.


JdeH:
Once again an important step!
Just curious, is the function `fix_time` the answer to the `now` accuracy problem?
Or does it do something else?

neriusmika:
Yes, it subtracts some delta if the time is near the next day or hour. Only problem can be with `today()` method if the tests will run at 23:59 :) If there will be problems, I think we can comment out those functions.

JdeH:
Well, lets see this as a "beauty spot". Anyone committing at 23:59 should just try again a little later...

PR #451, opened 2018-01-23
Decorator fix

neriusmika:
Fixed decorator name checking when importing from another module with parameters:

```python

@a_module.b_module.my_decorator('param1')
def my_function():
    ...
```


JdeH:
Once again much attention to (even typographic) details to fit in with the rest of the code.
Thanks Nerius!

PR #452, opened 2018-01-23
issubclass implementation

neriusmika:
Also fixed isinstance bug when checking primitives with `object` argument.

### DIFF
--- a/transcrypt/development/automated_tests/transcrypt/classes/__init__.py
+++ b/transcrypt/development/automated_tests/transcrypt/classes/__init__.py
@@ -65,17 +65,24 @@ def run (autoTester):
     autoTester.check (hasattr (a, 'p'))
     autoTester.check (hasattr (a, 'q'))
     
-    autoTester.check ('<br><br>Augmented isinstance<br>')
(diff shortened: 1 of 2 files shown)


JdeH:
Hi Nerius,

This will work, still two points of constructive criticism:

1. Function `isA` is in the global namespace, which I think we should avoid.
If `isA` has to be in the global namespace, could you use the name convention for system methods,
naming it `__isa__`. Developers will know that anything between `__ __` is reserved.

2. There's a lot of code duplication beween `isinstance` and `issubclass`.
I expect `isinstance` to be much smaller if it uses `issubclass`.
This few lines may not be a big deal, but accumulated over all built-in functions it will still amount to a significant amount of code, and guarding download size is very important as Transcrypt gets expanded over time.
A secundary advantage of this approach is that if `isinstance` achives its goal by calling `issubclass`, function `isA` can probably be local to `issubclass`, avoiding the name problem altogether.

neriusmika:
Ok, I'll try to remove code duplication.

neriusmika:
@JdeH, check the code now.

JdeH:
Fine!

PR #504, opened 2018-03-16
Do not use shlex.split under Windows

metamarcdw:
The xtrans pragma currently uses subprocess/shlex for running external transpilers.

Using shlex.split for subprocess under Windows is broken.
Luckily, it is also not necessary in that case.
https://stackoverflow.com/a/35900070


JdeH:
Thanks!

Issue #506, opened 2018-03-23
Command line switch for defining xtrans pragma shell command?


metamarcdw:
Since it is unlikely that any specific project would need to use more than one external transpiler, we could include as an option a command line switch for defining the shell command to be used in xtrans calls, rather than defining it in the call each time.

Would be happy to work on a PR if it seems like a good idea to everyone.

PR #507, opened 2018-03-24
Command line switch to define the shell command to be used in xtrans calls.

metamarcdw:
A command line switch can be used rather than defining it in the call each time.
Resolves issue #506


JdeH:
Merged with small changes afterward:

-xx changed to -xtr, since -xx prefix is reserved as prefix for yet another extended set of options, and to avoid having to rename -xtrans to -xxtrans (in line with other names).

- Line continuation in code of xtrans pragma removed, since it's the only place where it's used. The 80 chars per line restriction of PEP doesn't match current display sizes anymore.

Thanks for the contribution!

metamarcdw:
Thank you!

PR #513, opened 2018-04-14
[WIP]  Re module

GULPF:
I decided to rewrite the py2js regex translator from scratch, because my original code was not very good. It's nearly done, but needs some more tests , bug fixes, and cleanup.

Where should the documentation for the re module go? There are some limitations and Transcrypt specific features (notably, the option to use a JS-regex directly instead of using the translation step)  that the user should know about.


JdeH:
Since all tests run properly, I've allready merged this part of your work.

About the documentation, in the docs/sphinx directory, please add a file module_re.rst containing the docs for the re module.
I will add (or you can add) a chapter 10 to the docs, named `Libraries that come with the distribution` and refer to `module_re.rst` as a paragraph in this chapter, titled `Regular expressions: re`.

Issue #547, opened 2018-08-31
Missing stdlib function: unicodedata.normalize


eddieantonio:
The entire [`unicodedata`](https://docs.python.org/3.7/library/unicodedata.html) module is missing; this includes one of the most useful functions in this module, [`unicodedata.normalize()`](https://docs.python.org/3.7/library/unicodedata.html#unicodedata.normalize).

This function can be substituted with the ES6's [`String.prototype.normalize()`](http://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.normalize), if partial module definitions are okay.

JdeH:
Hi

I'd like to get an idea of what this entails. Apart from the normalize function, are there other parts of the _unicodedata_ that you often use? Can you indicate which ones are most relevant?

Partial module definitions are quite alright. In fact the whole of Transcrypt is designed with Parato's rule for IT in mind: 90% of the result is achieved with 10% of the effort. So what should be in the _unicodedata_ module to make it do the job in most cases?

KR
Jacques

eddieantonio:
It really depends upon the localization requirements for each project, but for me, the most useful members of `unicodedata` are:

1. [`unicodedata.name`](https://docs.python.org/3.7/library/unicodedata.html#unicodedata.name)
2. [`unicodedata.category`](https://docs.python.org/3.7/library/unicodedata.html#unicodedata.category)
3. [`unicodedata.lookup`](https://docs.python.org/3.7/library/unicodedata.html#unicodedata.lookup) — does the same thing as `\N{...}` escapes in strings, e.g., `"\N{FATHER CHRISTMAS}" == '🎅' == unicodedata.lookup('FATHER CHRISTMAS')`
4. [`unicodedata.numeric`](https://docs.python.org/3.7/library/unicodedata.html#unicodedata.numeric) and similar str -> numeric type conversions

I whipped up a quick proof-of-concept stub implementation of `unicodedata.normalize()`; I'll open a pull request soon.

PR #548, opened 2018-08-31
[WIP] Add unicodedata.normalize()

eddieantonio:
Let me know if there's a way I can add automated tests for this module!

Here's a simple test for this function:

```python
from unicodedata import normalize

# '\u00e8' == 'é'
# '\u0301' == 'e' + '\N{COMBINING ACUTE ACCENT}'
assert normalize('NFC', '\u00e9') == normalize('NFC', 'e\u0301'))
```

Addresses #547

### DIFF
--- a/transcrypt/modules/unicodedata/__init__.py
+++ b/transcrypt/modules/unicodedata/__init__.py
@@ -1,0 +1,9 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
+def normalize(form, unistr):
+    # Delegate to ES6's method:
+    # http://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.normalize
+    # Both ES6 and CPython require that `form` be the same four values.
+    # https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Modules/unicodedata.c
+    return String.prototype.normalize.call(unistr, form)


eddieantonio:
I'll whip up an autotest before this should be merged. Please let me know of any other things I should be aware of before this should be merged.

JdeH:
Thanks for this speedy contribution!

Currently Travis CI isn't up to par with Transcrypt. This will be a while, since the one who made the CI work currently isn't available, and I'm not exactly a Travis expert...

But you can run the autotests off-line. To add an autotest for this module, add a testlet to _../Transcrypt/transcrypt/development/automated_tests/transcrypt_. Also at an import and an autotester.run call to _autotest.py_. The principle of the autotest system is described at:

http://www.transcrypt.org/docs/html/autotesting_transcrypt.html

Due to the transition of TS 3.6 to TS 3.7 the documentation on autotesting may still lag back on some points. If so let me know so it can be updated. You should also be able to run the shipment test but it requires node.js' http-server to be installed (as the Python http server has some problems).

JdeH:
As you have contributed a new file, it can't break anything, so I'll already merge this.

JdeH:
I don't think there are other things you should be aware of.
I hope this module will grow according to the needs of people.

JdeH:
I've also manually drawn this into 3.7.
Are you still planning to add an autotest?

eddieantonio:
Yes! I'll whip up a PR within the next two days.

JdeH:
Thanks! And no hurry...

PR #549, opened 2018-09-04
Add autotests for module unicodedata.

eddieantonio:
Added the autotests for unicodedata.normalize(). See #548.

### DIFF
--- a/transcrypt/development/automated_tests/transcrypt/autotest.py
+++ b/transcrypt/development/automated_tests/transcrypt/autotest.py
@@ -43,6 +43,7 @@ if __pragma__ ('defined', 'undefined'):
 import module_datetime
 import module_itertools
 import module_math
+import module_unicodedata
 import modules
 import nonlocals
 import operator_overloading
@@ -98,6 +99,7 @@ if __pragma__ ('defined', 'undefined'):
 autoTester.run (module_datetime, 'module_datetime')
 autoTester.run (module_itertools, 'module_itertools')
 autoTester.run (module_math, 'module_math')
+autoTester.run (module_unicodedata, 'module_unicodedata')
 autoTester.run (modules, 'modules')
 autoTester.run (nonlocals, 'nonlocals')
 autoTester.run (operator_overloading, 'operator_overloading')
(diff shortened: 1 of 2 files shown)


JdeH:
Hi, thanks, that's fast...

I've already merged it, for me the simplest way to test it locally and it doesn't interfere with other functionality.
However I've temporarily outcommented the unicodedata test in ../transcrypt/autotest.py since I get the following error:

```
Error: cannot run autotest using CPython: 'charmap' codec can't encode characters in position 50-51: character maps to <undefined>
Traceback (most recent call last):
  File "D:\activ_tosh\geatec\transcrypt\qquick\Transcrypt\transcrypt\__main__.py", line 149, in main
    exec (sourceFile.read (), globals (), locals ())
  File "<string>", line 118, in <module>
  File "D:/activ_tosh/geatec/transcrypt/qquick/Transcrypt/transcrypt/modules\org\transcrypt\autotester\__init__.py", line 320, in done
    hg.generate_html(self.refDict)
  File "D:/activ_tosh/geatec/transcrypt/qquick/Transcrypt/transcrypt/modules\org\transcrypt\autotester\html.py", line 84, in generate_html
    dc.writeHiddenResults(f, refDict)
  File "D:/activ_tosh/geatec/transcrypt/qquick/Transcrypt/transcrypt/modules\org\transcrypt\autotester\html.py", line 173, in writeHiddenResults
    f.write ('<div id="{}">{}</div>\n\n'.format (refResultDivId, itemData))
  File "D:\python37\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 50-51: character maps to <undefined>
```

Probably this has something to do with my language settings on Windows 10. Since my language, dutch, has almost no non-ascii symbols, I have very little experience with unicode and its quirks. Do you have any idea what this error message is trying to tell me?

eddieantonio:
It seems that something in Transcrypt is writing files in your system's default encoding, [CP1252](https://en.wikipedia.org/wiki/Windows-1252). CP1252 tends to be the default encoding on Windows even in Canada (English/French), but in almost all circumstances, it is more useful to write files in UTF-8 as it can write _all_ Unicode characters and is ASCII-compatible.

The easiest thing to do is to somehow change your default system encoding to UTF-8. A better option may be to change lines in transcrypt that `open()` files and explicitly set the encoding to UTF-8.

JdeH:
In number of places in Transcrypt, files are explicitly opened as UTF-8.
But there may still be places where this isn't so.
I'll take another good look.

PR #560, opened 2018-09-19
Correctly recognize untouched files during compilation

phorward:
I found out that the compilation mechanism didn't work when trying to compile only changed source files. This should fix it for now.


JdeH:
Working on this, rooted rather deeply...

phorward:
> Working on this, rooted rather deeply...

Okidoki, this was just a solution that worked for me.

I've noticed that I did a mistake with the branches, so the code from pr #561 was also merged within this pull request, should I fix it here or will you work on an entirely different solution?

JdeH:
Fixed and commited, this ran quite deeply.
Feel free to open an issue it there are still problems with this.

PR #561, opened 2018-09-20
Fix for opov pragma on attribute function-calls

phorward:
Generates wrapping function with accumulator to avoid accidental double function calls.

Example Python:

```python
x(y).lower()
```

Resulted before patch (faulty code!) in:

```javascript
__call__ (__call__ (x, null, y).lower, __call__ (x, null, y));
```

Results now in:

```javascript
(function () {
      var __accu0__ = __call__ (x, null, y);
      return __call__ (__accu0__.lower, __accu0__);
}) ();
```


JdeH:
Thanks for correcting this!

PR #564, opened 2018-09-26
Use JavaScript Array.slice in '__getslice__' when applicable

daboross:
For step=1, the common case, performing a native array slice should be faster than copying the elements ourselves.

I'm not sure how to profile to tell whether it actually is, but it seems like it could be a good idea.

Let me know what you think, and whether there's anything else I can do in this PR. I'm not 100% familiar with the structure nowadays and I'm not sure whether anything needs to be done for pre-ES6 compatibility here?

### DIFF
--- a/transcrypt/modules/org/transcrypt/__builtin__.js
+++ b/transcrypt/modules/org/transcrypt/__builtin__.js
@@ -980,6 +980,10 @@ Array.prototype.__getslice__ = function (start, stop, step) {
         stop = this.length;
     }
 
+    if (step == 1) {
+        return Array.prototype.slice.call(this, start, stop);
+    }
+
     let result = list ([]);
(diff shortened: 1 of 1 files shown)


JdeH:
Hi

ES5 compatibility was dropped, somewhat anticipating on a bright future..., ES6 modules are needed anyhow for 3.7.
So I think this PR is OK.
Maybe a testcase can be added to the autotest (if it isn't there yet, I'm not sure).
I'll pull this one now already, since it matches the current head.

Glad to hear from you again!
Jacques

daboross:
Glad to contribute! I have a few patches that I realized I'd not actually made PRs for to check whether they'd be useful.

I can make tests if that would be helpful, I'm just not entirely sure how the testing framework works and haven't gotten the tests running on my computer yet.

PR #566, opened 2018-09-28
Remove generation of 'list()' around list literals

daboross:
This should be purely a output-code-niceness improvement.

`list()` currently just runs `Array.from`, so it seems redundant if we've already just created a fresh list using `[]` syntax. I don't believe it will have any affect other than just having another function run every time a list is instantiated.


JdeH:
It is indeed redundant.

PR #568, opened 2018-09-28
Make int() function match python's int() more fully

daboross:
Currently, Transcrypt's `int` allows float values and `int(4.4) = 4.4` (whereas `int(4.4)` in Python is `4`). Additionally, Python `int` supports an additional 'base' argument which is not supported in Transcrypt.

This should fix those two issues. However, it does mean a bit of duplication between `int` and `float` which I'm not sure how to remove.

This also tries to match python's behavior much more exactly, which might or might not be what we want. At minimum it would be nice if `int()` did floor values, but if we didn't want to support the rest this could just be `Math.floor (float (any))` rather than this whole deal.

### DIFF
--- a/transcrypt/modules/org/transcrypt/__builtin__.js
+++ b/transcrypt/modules/org/transcrypt/__builtin__.js
@@ -288,8 +288,21 @@ export function float (any) {
 float.__name__ = 'float';
 float.__bases__ = [object];
 
-export function int (any) {
-    return float (any) | 0
+export function int (any, radix) {
+    if (any === false) {
+        return 0;
+    } else if (any === true) {
+        return 1;
+    } else {
+        var number = parseInt(any, radix);
+        if (isNaN (number)) {
+            if (radix == undefined) {
+                radix = 10;
+            }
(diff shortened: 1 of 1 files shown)


JdeH:
It is not the amount of code that bothers me but rather the speed (or lack thereof).
It seems that parseInt (aFloat) will call toString on that float and then attempt to parse it as an int.
This seems like an extraordinary amount of processing.
But maybe some clever JavaScript VM will take a shortcut, I don't know...

I think the most frequent case is int (aFloat). In that case Math.floor (float (any)) would do fine.
Could it somehow be made to check that frequent case first and if that isn't "the case", attempt to convert from true/false or from string.
Or do you think float (any) will perform just as badly on a float as does parseInt?

daboross:
Sorry I've left this hanging! I'd completely forgotten about this.

I'll switch this to use `Math.floor(float(any))` once I get back working on these PRs. I had no idea `parseInt` would behave like that.

The radix behavior was mostly an addition for the sake of an addition - I don't have any use case for it - so I'm fine leaving that out.

PR #575, opened 2018-10-08
Extend allImportedNames to simple from...import statements

phorward:
Hi Jacques,

I've got the following, simple __init__.py in my library called `html5`:

```python
from html5.html5 import *
from html5 import ext
```

The current transcrypt master eats it, but only exports the names from `html5.html5`, not the `html5.ext`.

This was the faulty code before this fix:

```javascript
/* 000004 */ import * as ext from './html5.ext.js';
/* 000003 */ import {A, Abbr, Address, Area, Article, Aside, Audio...} from './html5.html5.js';
/* 000003 */ export {A, Abbr, Address, Area, Article, Aside, Audio...};
```

Now it is:
```javascript
/* 000004 */ import * as ext from './html5.ext.js';
/* 000003 */ import {A, Abbr, Address, Area, Article, Aside, Audio...} from './html5.html5.js';
/* 000003 */ export {ext, A, Abbr, Address, Area, Article, Aside, Audio...};
```

This pull requests fixes it for now and adds `ext` also to the allImportedNames set of the module.
Is this okay?


JdeH:
This is a correct pull request, thanks.
Since some other changes were made to the same segment of code, I'll manually add it to the module aliases branch.

Issue #552, opened 2018-09-07
with __pragma__(...):


kiwi0fruit:
It would be nice to have in addition to in-line comment pragmas the new more explicit pragma via `with` construction. It switches to pragma needed then switches back to the global value:

```py
with __pragma__(...):
    do()
more()
```

happy-monk:
Do you mind if I try to implement this?

JdeH:
@happy-monk 

It would be a nice addition if it doesn't complicate the compiler too much.
You're welcome to give it a try.
I've marked it 'under construction' for now.

Kind regards
Jacques

happy-monk:
The `with __pragma__` now works in most cases, but with a couple of issues.

* First of all, if you try to nest pragmas of the same kind, you will see rather confusing results. For example:

```
with __pragma__('opov'):
    a = b + c
    with __pragma__('opov'):
        x = y + z
    d = e + f
```

`a = b + c` and `x = y + z` will be affected by 'opov', but `d = e + f` will not.

This is because `with __pragma__` is implemented now by appending/prepending `__pragma__('xxx')` and `__pragma__('noxxx')` to the context code. So the nested `with` will always disable its effect instead of restoring the previous state.

The same applies to the single-line activation variety of pragmas, because it's implemented in the same way.

Of course, this is not kind of thing you should see in final/production/etc. code, but it surely will confuse people and get in the way of experimentation with code.

To solve this some architectural solution is needed. There are several way to handle this, but, I think, the easiest way would be to implement some kind of stack for translator flags and utilise it for `with` handling and single-line activation. The `__pragma__(push)` and `__pragma__(pop)` could also be realised for use with non-context pragmas.

JdeH:
I agree that this behavior will be confusing if you encounter it while experimenting.
Still I think that the 'with __pragma__' variant is quite useful.
So I'd like to include it in the upcoming release, but will pay attention to its limitations in the documentation.

PR #579, opened 2018-10-18
[WIP] __pragma__ as context manager

happy-monk:
Preliminary implementation of #552, placed here to get some feedback.

I've managed to make it work for the most of the pragmas, including 'skip' and 'alias'. The `__exit__` part of the `with` statement is implemented by prefixing `no` to the name of pragma (or removing it) for now, which should work well in most cases, but it will not always restore the flags to the same state they were before `with`. There is various ways to go about this problem.

I think, the simplest solution would be to write a pair of push/pop methods (and push/pop pragmas, probably) to save and restore all flags at once and use them before and after with-clause. This will have side effect of creating a scope for pragmas inside each with-clause, but the effect is probably not undesired.

### DIFF
--- a/transcrypt/modules/org/transcrypt/compiler.py
+++ b/transcrypt/modules/org/transcrypt/compiler.py
@@ -3469,6 +3469,7 @@ return list (selfFields).''' + comparatorName + '''(list (otherFields));
             self.prevTemp ('break')
 
     def visit_With (self, node):
+        from contextlib import contextmanager, ExitStack
         self.adaptLineNrString (node)
 
         @contextmanager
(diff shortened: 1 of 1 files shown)


JdeH:
Hi,

As you may have concluded, I haven't been able to work on Transcrypt for some months.
But now it's in my weekly planning again, so I pick up where I left.

I merged your solution for nested context managers, that solved a long standing bug.
If I am right this solution for using pragma's as context managers builds on that solution.

To provide you with feedback: I think pragma's as context managers are a good idea, so I'd like to incorporate those in Transcrypt. The code you wrote for this looks good.

I'd like to take it into account into the dev_fall_2019 branch, which is then later merged into master as a whole after thorough testing, to create a new release.

Switching the target from master to dev_fall_2019 resulted in a minor merge conflict, but I think that's easily solved, so no problem there.

You indicate that it's a Work In Progress. Does this mean that it's best to wait with merging this into dev_fall_2019 until you finished it?

Can you tell me which pragma's are still problematic?

Kind regards and thanks for the contribution "In Progress"!
Jacques

JdeH:
Once again this looks very clean and elegant!
Although it's still marked WIP in its title, I'll merge it into the dev_fall_2019 branch, to prevent conflicts emerging as sideeffect of other modifications.

happy-monk:
Sorry, I was too busy with work for any meaningful answer. I'll try to write a comment in the main issue about the remaing problems in a few days.

PR #583, opened 2018-10-28
Fix `with` statement for multiple context managers

happy-monk:
The previous implementation would call the body of `with` multiple times, if it contains multiple context managers.

For example, for the following code:

```
  with a, b:
      fn()
```

would be generated the equivalent of this:

```
  with a:
      fn()

  with b:
      fn()
```


JdeH:
I think solving code generation for nested context managers with nested context managers in the compiler is a very elegant solution!
Compliments, I'll merge this into the development branch.
Could you supply some tests cases for the autotest?
If not I'll try to do that myself.

PR #584, opened 2018-10-30
Corrected sourcePath info in Errors

phorward:
Currently, Transcrypt reports something like
```
Error while compiling (offending file last):
        File '<unknown>', line 32, at import of:
        File '<unknown>', line 4, at import of:
        File '<org.transcrypt.compiler.Module object at 0x7f89eb64fda0>', line 4, namely:
```
this fixes it so that
```
Error while compiling (offending file last):
        File '/home/user/game/game.py', line 32, at import of:
        File '/home/user/game/html5/__init__.py', line 4, at import of:
        File '<org.transcrypt.compiler.Module object at 0x7fd6746f6c50>', line 4, namely:
```
will be reported.

### DIFF
--- a/transcrypt/modules/org/transcrypt/utils.py
+++ b/transcrypt/modules/org/transcrypt/utils.py
@@ -195,13 +195,13 @@ class Error (Exception):
             
     def __str__ (self):
         result = 'Error while compiling (offending file last):'
-        
-        try:
-            sourcePath = importRecord [0] .sourcePath
-        except:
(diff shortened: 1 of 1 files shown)


JdeH:
Thanks for finding the cause of this long standing bug.
It must have gotten broken somewhere along the way.
It just never got on top of my priority list, but it was very annoying in practical use!

JdeH:
There are more (annoying but fixable) problems with error reporting.
It's being worked on right now.

Issue #587, opened 2018-11-05
use __getattr__ and return self doesn't allow method chaining


pjbonestroo:
I use `__getattr__` and `__setattr__` to wrap an html element. Then I'm able to set properties on the html element using the wrapper. For this example, the html element is replaced by ` class Element` to let it work in CPython.

It all works nice, but only the method chaining doesn't work:

```
class Element():
    def __init__(self):
        self.message = "Goodbye"

    def sayBye(self):
        print(self.message)

class Wrapper():
    def __init__ (self, element):
        self.element = element

    def __setattr__ (self, name, value):
        """ set attribute on element if it already has the attribute """
        if name != "element" and hasattr(self.element, name):
            setattr(self.element, name, value)
        else:
            self.__dict__[name] = value

    def __getattr__ (self, name):
        """ get attribute from element if this object doesn't have the attribute """
        result = getattr(self.element, name)
        # if result is a function, bind self.element to it
        if hasattr(result, 'call') and hasattr(result, 'bind'):
            result = result.bind(self.element)
        return result

    def sayHello(self):
        print("Hello")
        return self


e = Element()
w = Wrapper(e)

#
# Usage
#

w.message = "Bye" # set attributes on e, using w
e.sayBye()
w.sayBye() # call functions on e, using w

# and method chaining should work:
w.sayHello().sayBye()
```

Returning self in method `sayHello` gives the `Wrapper` object, but not the javascript Proxy object. Is this intentional, and thus a restriction (it works in CPython)?
Or is it a bug?

pjbonestroo:
Hi Jacques,
I hope it can be done while maintaining performance. I didn't think about a solution yet. If I find time, I will look into it.
Thanks in advance for helping.
Regards, Pieter Jan

pjbonestroo:
A trial in the generated javascript would be like:

org.transcrypt.__runtime__.js
```
export function __get__ (self, func, quotedFuncName) {
    if (self) {
        if (self.hasOwnProperty ('__class__') || typeof self == 'string' || self instanceof String) {
            if (quotedFuncName) {
                Object.defineProperty (self, quotedFuncName, {
                    value: function () {
                        var args = [] .slice.apply (arguments);
                        return func.apply (null, [self] .concat (args));
                    },
                    writable: true,
                    enumerable: true,
                    configurable: true
                });
            }
            return function () {
                var args = [] .slice.apply (arguments);
                if (self.__proxy__ != undefined){
[...]

PR #591, opened 2018-11-07
Added __proxy__ to instance when javascript Proxy object is created.

pjbonestroo:
Solves issue #587


JdeH:
Hi Pieter Jan,

It has been a while... I hope you're well.
I've (finally) manually merged this in the dev_fall_2019 branch.
Nice solution with very little overhead.
I've tried to avoid the calltime test and make it instantiation time, but I failed at that.
Also added an autotest based on your use case / example to the transcrypt/proxies testlet.

Kind regards
Jacques

PS In case of remaining problems feel free to reopen.

PR #654, opened 2019-07-20
Travis CI: The sudo tag is now deprecated in Travis

cclauss:
__sudo: required__ no longer is.  The __sudo__ command is now always available in Travis CI and there is no way to turn it off.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 dist: xenial
-sudo: yes
 
 addons:
     chrome: stable


JdeH:
@cclaus
Pulled this to the wrong branch (master).
Reverted and tried to pull to the right branch (dev_fall_2019) but GitHub refused to restore the original pull request.
So redid it by manual edit in the dev_fall_2019 branch.
Will be merged with master after completion of some other issues.
Thanks for the contribution!

PR #673, opened 2019-09-08
Travis CI: Run parallel test on Python 3.8

cclauss:
Also, __Xenial__ is now Travis CI's default distro.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-dist: xenial
-
 addons:
     chrome: stable
     apt:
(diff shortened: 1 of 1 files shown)


JdeH:
Thanks for the contribution!

There's still some incompatibility between Transcrypt and Python 3.8.
It seems that the syntax tree is a little different, so I had to drop certain parts of the autotest.
But this will be repairable.
Meanwhile it's a good thing to already (partially) test on Python 3.8.
Some Python 3.8 features, like the walrus operator, will most likely be added to Transcrypt 3.8.
Since the Python parser is used, the compiler will require Python 3.8.

Kind regards
Jacques

PR #686, opened 2019-10-23
React Demo with hooks

happy-monk:
Since React proposes [hooks](https://reactjs.org/docs/hooks-intro.html) as the preferred way of writing new components, I've tried to rewrite React Demo using hooks. This looks considerably cleaner than createClass-based implementation.

Also I took liberty of creating `org.reactjs` package with some additions to allow the use of hooks as decorators.

### DIFF
--- a/transcrypt/demos/react_demo/react_demo.html
+++ b/transcrypt/demos/react_demo/react_demo.html
@@ -3,8 +3,8 @@
   <head>
     <meta charset="UTF-8" />
     <title>Hello React!</title>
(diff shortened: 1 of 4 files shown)


JdeH:
I must say that, coming from C++, I have found JS classes always a bit weird and off the mark.

Following your link, I've read that the React team promoted the use of hooks over classes, e.g. because the use of the `this` reference in JS confuses people (I can relate to that).

I have zero practical React experience, but since this seems to be the direction React is taking, I think Transcrypt should follow.

Thanks for the contribution!

PR #702, opened 2020-01-09
Fix slice to handle negative steps

### LINKED ISSUE #455: Reverse string slice goes into infinite loop
How to reproduce:

```javascript
"abc".__getslice__(null, null, -1)
```
This goes on forever.

Basically what I'm trying to do is this:

```python
assert "abc"[::-1] == "cba"
```

pjz:
This fixes #455 to, as far as I can tell, work correctly.


JennaSys:
Accepted PR but unable to auto-merge, so added changes manually into dev_3.9.4 branch.

Issue #588, opened 2018-11-05
Getting JavaScript TypeError on None-value parameter check


phorward:
Hi!

Given the following piece of code
```python
def domCreateElement(tag, ns=None):
	"""
		Creates a new HTML/SVG/... tag
		  :param ns: the namespace. Default: HTML. Possible values: HTML, SVG, XBL, XUL
    """
	uri = None

	if ns == "SVG":
		uri = "http://www.w3.org/2000/svg"
	elif ns == "XBL":
		uri = "http://www.mozilla.org/xbl"
	elif ns == "XUL":
		uri = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"

	if uri:
		return document.createElement(uri, tag)

	return document.createElement(tag)
```
I'm getting this error in JavaScript with the latest Transcrypt 3.7.11, but also before:
```
 [Show/hide message details.] TypeError: right-hand side of 'in' should be an object, got null[Learn More] org.transcrypt.__runtime__.js:1541:22
__eq__
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:1541:22
domCreateElement
http://localhost:8000/__target__/html5.core.js:74:26
__call__
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:166:29
get __init__/<
http://localhost:8000/__target__/html5.core.js:601:32
__call__
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:166:29
get __init__/<
http://localhost:8000/__target__/html5.core.js:5563:16
__call__
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:166:29
get __init__/<
http://localhost:8000/__target__/game.js:23:16
__new__
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:149:22
cls
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:92:33
__call__
http://localhost:8000/__target__/org.transcrypt.__runtime__.js:166:29
<anonymous>
http://localhost:8000/__target__/game.js:98:67
<anonymous>
http://localhost:8000/__target__/game.js:93:16
InnerModuleEvaluation self-hosted:4355:5 InnerModuleEvaluation self-hosted:4342:17 evaluation self-hosted:4301:9 
```
The reason for this is, that the `if ns == "SVG":` is causing the code 
```javascript
if (__t__ (__eq__ (ns, 'SVG'))) {
[...]

PR #734, opened 2020-05-24
Check nullity for __eq__ and __ne__

pfeodrippe:
`__eq__` for `Array` checks for the lenght of `b`, but `b` could be `null`

### DIFF
--- a/transcrypt/modules/org/transcrypt/__builtin__.js
+++ b/transcrypt/modules/org/transcrypt/__builtin__.js
@@ -2104,7 +2104,7 @@ export function __and__ (a, b) {
 // Overloaded binary compare
 
 export function __eq__ (a, b) {
-    if (typeof a == 'object' && '__eq__' in a) {
+    if (b != null && typeof a == 'object' && '__eq__' in a) {
         return a.__eq__ (b);
     }
     else {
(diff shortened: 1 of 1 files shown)


mentalisttraceur:
I do not understand the reason for this special case. Should objects with custom equality checks not be allowed to decide if they are equal to `null` (and `undefined`)?

mentalisttraceur:
What problems are caused by permitting `null` (or `undefined`) to pass through in here? Do they outweigh the possible benefits of being able to have a class which compares equal to them?

mentalisttraceur:
These are genuine questions: I don't know, I haven't done enough work with Transcrypt to say.

But my intuition is that classes should be able to receive raw `null` and `undefined` in their custom comparison methods, because this is more flexible and obvious to me than a special case where those specific values bypass custom comparison logic.

mentalisttraceur:
Also, possibly related: independent of this change, I don't understand why these built-in `__eq__` and `__ne__` functions are one-sided (they have a case for calling `a.__eq__(b)` but not a symmetric case for calling `b.__eq__(a)`).

Couple possibilities:

1. That's not implemented at all. If class A implements `__eq__` and you do a comparison like `native_js_object == instance_of_A`, Transcrypt never tries `instance_of_A.__eq__(native_js_object)`. That seems like such a profound gap in functionality, that I don't think that's likely to be the case.

2. That's implemented inside the `.__eq__` and `.__ne__` that are on the objects. If so, then I would expect an `if(typeof b == 'object' && '__eq__' in b)` check to happen in those methods before trying `b.__eq__(a)`, and that should protect against any inherent problem with letting `null` or `undefined` fall through.

JennaSys:
Maybe related to #588

PR #765, opened 2020-12-19
Add time.monotonic

mentalisttraceur:
Added `time.monotonic` - trivial one-line implementation based on `performance.now`.

The docstring is a simplified and clarified version of the official Python documentation.

### DIFF
--- a/transcrypt/modules/time/__init__.py
+++ b/transcrypt/modules/time/__init__.py
@@ -244,6 +244,17 @@ def time():
     return Date.now() / 1000
 
 
+def monotonic():
+    """monotonic() -> float
+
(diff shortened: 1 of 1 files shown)


mentalisttraceur:
I can imagine a good test being:

1. Run the `time.monotonic` function, save the result,
2. schedule a timeout (or interval!) callback for `n` seconds in the future, and
3. in the callback:
    1. run the function again in the callback,
    2. subtract the previous result from the new result, and
    3. that it's approximately `n` (like `difference >= n * 0.95 || difference <= n * 1.05`).

But I didn't see any tests like that already, and I might not have time to implement such a test any time soon, so I didn't want to start writing it or hold up the pull request until I did unless it's a blocker/required.

PR #777, opened 2021-06-07
Add a simple json module to Transcrypt

nandoflorestan:
Transcrypt is awesome, thank you!

I want to use a single validation library on both sides of the wire.  I have been trying to make a few Python validation libraries to work in the client, too. This can solve the problem of keeping validation code in sync.  But in order to do so, I need the json module.

Most Python validation libraries use the json module.  It's probably impossible to implement the subtleties of the json module, but 99% of its usage is simple, and this we can do in very little code.

Let me know about any problems you see, of course.


JennaSys:
May be related to #868

PR #778, opened 2021-06-07
Please help with defaultdict implementation

nandoflorestan:
While trying to make a Python validation library work in the browser, I thought I'd prefer not to change the library and just pay for it with 25 lines in Transcrypt.

I thought this simple implementation of defaultdict would work, but Firefox complains:

    Uncaught TypeError: meta is undefined
    __class__ org.transcrypt.__runtime__.py:4
    <anonymous> collections.py:4

I suppose *dict* is not a Python class in Transcrypt -- is it just a JS object?

In any case, please let me know what's going on and give me a north.

### DIFF
--- a/transcrypt/modules/collections.py
+++ b/transcrypt/modules/collections.py
@@ -1,0 +1,28 @@
+"""Python's collections module -- for Transcrypt."""
+
+
+class defaultdict(dict):
+    """Dictionary that takes a factory parameter and always returns a value."""
+
+    def __init__(self, default_factory=None, *args, **kwargs):  # noqa
+        if not callable(default_factory) and default_factory is not None:
+            raise TypeError("first argument must be callable or None")
+        super().__init__(*args, **kwargs)
+        self.default_factory = default_factory
+
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


JennaSys:
I think I've seen it mentioned in the Transcrypt docs somewhere, but on the JavaScript side, dict is just a JS object with extra properties and methods tacked on to it.  Depending on what's happening on the JS side, these extra methods and properties can get "lost" and the JS pseudo-dict gets turned back into a plain JS object.  I have this happen a lot when the object is being passed in to a function as a parameter (such as in the case with React props). I'm not sure if this is related to what is happening in your case or not.

On the Python side, if you are just directly accessing the dict elements it works fine anyway. But if you need to use dict methods, I've worked around it by adding a dict() cast before trying to use the object on the Python side.  This forces the methods and properties to be appended to the JS object right before you need to call them.

nandoflorestan:
@JennaSys 

Thank you very much for helping out.  I don't think that is the issue here because the Firefox exception happens when I try to instantiate a defaultdict -- it might be that Transcrypt is not prepared to let anyone subclass dict.

The game continues...

Issue #699, opened 2019-12-11
`time.monotonic` using `performance.now`?


mentalisttraceur:
Monotonic clocks are the best practice time source for any code that cares about how much time has passed between moments.

This looks like it would be trivial to add to the `time` module:

```python
def monotonic():
    return performance.now() / 1000
```

Advantage of doing this in Transcrypt's library is that Python code already written to use or prefer `time.monotonic` will just work.

mentalisttraceur:
If I don't hear a "no" maybe I'll just get a PR going once I have time to set up and test it, and we can discuss it in more depth then.

This one should be easier for me to get started on than the other change I suggested months ago.

JdeH:
Sorry this issue has been without an answer for so long.
A contribution to this regard is appreciated!

mentalisttraceur:
@JdeH No problem! I've submitted pull request #765 with it now.

JennaSys:
Implemented in v3.9.2

PR #787, opened 2021-07-23
Add time.monotonic

mentalisttraceur:
Added `time.monotonic` as suggested in #699 - it is a trivial one-line implementation based on `performance.now`.

The docstring is a simplified version of the official Python documentation.

### DIFF
--- a/transcrypt/modules/time/__init__.py
+++ b/transcrypt/modules/time/__init__.py
@@ -244,6 +244,17 @@ def time():
     return Date.now() / 1000
 
 
+def monotonic():
+    """monotonic() -> float
+
+    Return the value (in seconds) of a monotonic clock - a clock that
+    cannot go backwards. The reference point of the returned value
+    is undefined, so only the difference between the results of
+    consecutive calls has guaranteed meaning.
+    """
+    return performance.now() / 1000
+
+
 def asctime(t):
     return strftime('%a %b %d %H:%M:%S %Y', t)
 


mentalisttraceur:
P.S. My apologies for remaking #765 - I wanted to be proactively helpful and update this to fix the spurious merge conflict that got introduced by other changes at. But since I had deleted the fork from my account (I don't like having forks sitting around on my GitHub), recreating the pull request seemed to be the only way for me to do so.

But to be clear, I still really don't want forks sitting around on my GitHub account, and after not getting any response on this trivial PR for seven months, I'm not feeling very optimistic or patient about it being worth it to do so, so we're now in the same state: the fork is already gone from my account and thus I can't update this PR.

The difference this time is that I won't be recreating a third time, unless I get some explicit signal that the help is actually wanted.

mentalisttraceur:
Every once in a while I think "maybe it _is_ worth it to have one or more tests for this", but what, exactly, would such tests be proving?

We don't want to be testing the behavior of `performance.now()` itself, because if there's a bug in that, that's not Transcrypt's problem - whoever implemented `performance.now` needs to be testing that.

I presume we don't want Transcrypt to be in the business of implementing backwards-compatible polyfils for `performance.now` which try to degrade while at least maintaining monotonicity on browsers that don't have it, or workarounds for buggy implementations of `performance.now` if there is any browser that requires that, and I don't know of any changes in browser APIs on the horizon which might make us use a more complex implementation.

I also don't think Transcrypt wants to be in the business of supporting old browsers which don't have `performance.now` at all. But even if we do, this implementation doesn't even prevent importing the `time` module on such browsers - it's just that if you call `time.monotonic()` you'll get the same exception as you would if you tried calling `performance.now()`.

So what exactly is the testable surface area? That we didn't make a typo? That someone didn't accidentally delete or add a `0` in the `/ 1000`? That sounds silly, but I _do_ see the value of that - it's a good safety railing to have against accidents. So we could do a one-liner test
verifying that `time.monotonic` returns a number, and if that number seem to be approximately the output of `performance.now() / 1000`:

```python
int(time.monotonic()) <= int(performance.now() / 1000)
```

I didn't initially put that in because it seems like it's almost redundant, and it takes me more work to verify that the test is appropriately scoped and correct than to just verify the code itself (much more work, if you have the attention to detail to notice that the above test is only going to be guaranteed correct if the left side of `<=` is guaranteed to be evaluated before the right side, and the standard of care to go look up the relevant ECMAScript standard to verify if that actually is guaranteed).

But I will gladly go ahead and add So if I get an explicit statement that this is wanted. And I'm open to other suggestions about what's worth testing. Maybe I'm missing something  in my above reasoning.

That's why my initial description of #765 that I wrote back in December 2020 said

> Please advise if you want any tests for it!

mentalisttraceur:
> and it takes more work to verify that the test is appropriately scoped and correct than to just verify the code itself

Case in point:

I don't remember if `<=` is *guaranteed* in JavaScript to evaluate the left operand before the right operand.

Knowing me, I *probably* checked that when I first wrote that example test line. But right now I don't remember. And if it's not guaranteed, then that test is just wrong/unreliable/nondeterministic.

In the amount of time it even took me to remember this possibility, I can validate that this `time.performance` is still implemented correctly. In the amount of time it would take me to describe it, or in the amount of time it would take me to look up relevant ordering guarantees in JavaScript, I could do that several times over.

JennaSys:
Accepted but unable to auto-merge, so merged manually.

mentalisttraceur:
Glad to see this finally make it in! Thank you!

PR #834, opened 2022-12-04
Embedded HTML similar to JSX

tomkcook:
This PR currently depends on a development version of aroberge/ideas - see https://github.com/aroberge/ideas/pull/42 for the latest version.

This adds HTML embedded in Python, between delimiters `>>|` and `|<<`.  HTML elements are converted into a function call to `el(element_name, attrs_dict, children_list)`.  So this function:

```
def some_html():
    return >>|
        <div class="div-class">
            Some Text
            <a href="http://google.com">
                A Link
            </a>
        </div>
    |<<
```

is rendered as:

```
def some_html():
    return el("div", {"class": "div-class"}, [
        "Some Text",
        el("a", {"href": "http://google.com"}, [
            "A Link"
        ])
    ])
```

The `el()` function is not provided and must be already available in the context where the HTML appears.  Its signature is intended to enable this use:

```
react = require("react")
el = react.createElement
```

I'm posting this PR to get feedback; if there is a possibility it will be merged, please say so and I will add some unit tests etc and get it into proper shape.  If there is no interest in this sort of PR, please let me know.

My apologies for a relatively large number of whitespace changes.  My editor regularised a number of line endings and removed whitespace at the end of lines.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,13 +16,14 @@ setup (
 		read ('README.rst')
 	),
 	keywords = ['python', 'javascript', 'transpiler', 'compiler', 'browser', 'web', 'multiple inheritance', 'transcrypt', 'django'],
-	url = 'http://www.transcrypt.org',	
+	url = 'http://www.transcrypt.org',
 	license = 'Apache 2.0',
 	author = 'Jacques de Hooge',
 	author_email = 'jacques.de.hooge@qquick.org',
 	packages = ['transcrypt'],
 	install_requires = [
-		'mypy'
+		'mypy',
+		'ideas',
 	],
 	entry_points = {
 		'console_scripts': [
--- a/transcrypt/__main__.py
+++ b/transcrypt/__main__.py
@@ -31,7 +31,7 @@ modulesDir = '{}/modules'.format (transpilerDir)                                
 sys.path = [item.replace ('\\', '/') for item in sys.path]
 
 # Unload org from a packages dir, if it happens to be there in the CPython installation
-sys.modules.pop ('org', None)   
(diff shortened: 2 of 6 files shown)


JennaSys:
I appreciate the PR submission and what you are trying to accomplish with it, but will unfortunately have to decline it as it is adding a specific use case to Transcrypt which is intended to remain more use-case agnostic. The added dependency on the [ideas](https://github.com/aroberge/ideas) library is also not ideal.

Have you looked at using the Transcrypt [xtrans compiler directive](https://www.transcrypt.org/docs/html/special_facilities.html#using-an-external-transpiler-pragma-xtrans-translator-cwd-workingdirectory) to shell out to an external transpiler for the JSX? I experimented with this myself in the past and got it working on Linux, but had issues with getting it to work on Windows for some reason. If you are interested in this approach, start a new topic in the [discussion area](https://github.com/QQuick/Transcrypt/discussions) and I'll see if I can dig up the code I used for you.

Since I also use Transcrypt with React myself, I have a number of opinions/ideas on how to work with it and would welcome discussions on that topic as well. One thing in particular that I did was to create a decorator so that I didn't need to use the `el()`s all over the place:
```python
def component(react_component):
    def react_element(props, *children):
        return React.createElement(react_component, props, *children)

    return react_element
```
This simple [React tutorial demo](https://github.com/JennaSys/tictacreact2) shows  it in use, and I also did a [write-up](https://dev.to/jennasys/no-el-eliminate-explicit-calls-to-createelement-when-using-python-to-code-react-applications-5214) on it a while back.

PR #879, opened 2024-08-23
Prevent polluting prototypes with enumerable props

### LINKED ISSUE #816: Prototype pollution
After importing a transcrypt module, various prototypes such as Array, String, Uint8Array are polluted with python specific helper functions. These appear to be coming from the org.transcrypt.__runtime__.js module such as

```
Array.prototype.extend = function (aList) {
    this.push.apply (this, aList);
};
```

Unfortunately these functions are listed as enumerable resulting in unexpected behaviour in external code ran later when its iterating even when the array wasn't created in the python code. Executing the following in the js console after simply importing the module

 ```
for( v in []){
    console.log(v)
}
```

results in all these items being output
```
_class__
__iter__
__getslice__
__setslice__
__repr__
__str__
append
py_clear
extend
insert
remove
index
py_pop
py_sort
__add__
__mul__
__rmul__
__bindexOf__
add
discard
isdisjoint
issuperset
issubset
union
intersection
difference
symmetric_difference
py_update
__eq__
__ne__
__le__
__ge__
__lt__
__gt__
```

Changing it to be defined as the following appears to resolve the issue and doesn't appear to impact the functionality in my test case
```
Object.defineProperty(Array.prototype, 'extend', {
	value: function (aList) {
		this.push.apply (this, aList);
	},
	enumerable: false,
	writable: false
})
```

### PR DESCRIPTION
**Pull request by JGreenlee**
Fixes https://github.com/TranscryptOrg/Transcrypt/issues/816

Assigning functions directly to the prototypes of Number, Array, String, etc., causes them to be enumerable, meaning they can show up in "for .. in .." loops and cause unexpected behavior!

Instead we can use `Object.defineProperty` (for which there is already a helper called `__setproperty__`). With this method, i) new props are non-enumerable by default, and ii) we can avoid unnecessarily reassigning the same methods (in case there are multiple instances of Transcrypt running).

### DIFF
--- a/transcrypt/modules/org/transcrypt/__builtin__.js
+++ b/transcrypt/modules/org/transcrypt/__builtin__.js
@@ -1925,8 +1925,14 @@ _dictmethods.__setitem__ = function (aKey, aValue) {
     this [aKey] = aValue;
 }
 
+__setproperty__ (dict.prototype, '__class__', { value: dict, writable: true });
+// Assign 'dict' methods to dict.prototype
+for (let [key, value] of Object.entries (_dictmethods)) {
+    __setproperty__ (dict.prototype, key, { value });
+}
+
 export function dict (objectOrPairs) {
-    var instance = {};
+    var instance = Object.create (dict.prototype);
     if (!objectOrPairs || objectOrPairs instanceof Array) { // It's undefined or an array of pairs
         if (objectOrPairs) {
             for (var index = 0; index < objectOrPairs.length; index++) {
@@ -1968,7 +1974,7 @@ export function dict (objectOrPairs) {
             }
         } else if (objectOrPairs instanceof Object) {
             // Passed object is a JavaScript object but not yet a dict, don't copy it
-            instance = objectOrPairs;
+            Object.assign (instance, objectOrPairs);
         } else {
             // We have already covered Array so this indicates
             // that the passed object is not a js object - i.e.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by JGreenlee** [comment]
### Put dict methods on the dict.prototype

I noticed that `list` methods get assigned to `Array.prototype`, `str` methods get assigned to `String.prototype`, etc.
`dict` methods do not get assigned to `Object.prototype` (I am guessing this is for a good reason)

The current code has the `dict` methods being added to each new instance in the constructor; but it is tidier to declare them on the `dict` prototype and then create each dict instance from that.
This is also more efficient because properties placed on the dict prototype need only be declared once, as opposed to being declared for each dict.

**Comment by JGreenlee** [comment]
The properties & prototype chain for an example dict `foo`:

<table>
<td>

Before (on `master`):

image

</td>
<td>

After changes

image

</td>

**Comment by JennaSys** [comment]
This is one I've had an eye on to tackle at some point sooner than later, so thanks for looking into it.  I can't dig into it right at this moment, but are there any possible unintended side effects by changing this that you can think of?

**Comment by JGreenlee** [comment]
If for some reason, someone was expecting the methods of str or list to be included while iterating over an Array or String, they would no longer see those.

```js
for (let e in []) {
  console.log(e)
}
// on master, prints all the methods of py `list`
// after this PR, prints nothing
```

I don't know why someone would be relying on that behavior though.

PR #900, opened 2025-06-22
Fixed many python compatiblity issues

kochelmonster:
- relative import
- metaclass
- descriptor protocol
- id, frozenset, getattr
- missing modules
- type and super


JennaSys:
Thank you for the contribution. There is a lot in here that I'll have to dig through. TBH, I'll likely end up having to cherry pick some of the updates rather than do one massive merge. But, at least based on the commit messages, I'm looking forward to adding some of these features.

Also just so you know, I'm currently in the process of updating the Transcrypt compiler to be compatible with Python 3.13 and any new features will need to be added after that is released.

